PR #1, opened 2021-05-18
NpcBalloon.cpp

farisawan-2000:
List of changes:
 - add diff tools (including ppc diff script)
   - this uses `powerpc-linux-gnu-objdump`, so definitely let me know if that needs to change
 - Decompile TNpcBalloon class
 -  Move stuff into headers and a bit of naming (let me know if that needs changing too)


QbeRoot:
First C++ PR hype!

Would it be possible to make the diff tools work with devkitPPC's `powerpc-eabi-objdump` instead?

farisawan-2000:
Should be good to go now; My implementation might need review though (it's all in diff_settings.py)

PR #41, opened 2023-09-22
Add Docker build enviroment

breadbored:
This change was made because the Mac environment can be problematic, especially with Wine. This is just a simple addition to make getting started with the project easier.

I also updated the `GC_WII_COMPILERS.zip` now that the link is dead. I linked to the archive.org snapshot of the file.


EpochFlame:
[the zip link](https://cdn.discordapp.com/attachments/727918646525165659/1129759991696457728/GC_WII_COMPILERS.zip) 

^ this is the updated compiler archive

breadbored:
Got it, that link works for me. I updated the link in the latest commit

PR #49, opened 2025-02-26
Decompiling TWatergun

TheAzack9:
This is still WIP, just marking progress


include/Player/WaterGun.hpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    u8 getJointIndex(int index, int jointIndex) const {
+        return mNozzles[index].mJoints[jointIndex].jointIndex;
+    }
+    
+    NozzleData mNozzles[6];
+};
+// Define the global variable in .data section
+const static TNozzleBmdData nozzleBmdData = {
```
Mrkol: This is probably wrong, because with a `static` variable in a header this symbol will appear in every single TU that the header is included into. Instead it probably should be `extern const TNozzleBmdData nozzleBmdData;` (the declaration) in the header and a regular non-static definition in the .cpp file with the initializer

include/Player/WaterGun.hpp
```diff
@@ -0,0 +1,285 @@
...
+            nullptr,                                    // mHelmetPath
+            "/mario/watergun2/normal_wg",              // mPath
+            "/mario/watergun2/normal_wg/normal_wg.bmd", // mBmdPath
+            1,                                          // mNumEmitters
+            2,                                          // _15
+            {
+                {1, 0},
+                {4, 0},
+                {4, 0}
+            }
```
Mrkol: JFYI: before merging you'll have to format, and to make clang-format not uglify large initializers like this one what you need to do is to leave an oxford comma after the last element in each subobject

include/Player/WaterGun.hpp
```diff
@@ -0,0 +1,285 @@
...
+    inline void setEmitPt() {
+        int index;
+        int offset;
+        // TODO: Register r27 and r28 are swapped on some instructions. I cannot figure out how to swap them correctly.
+        for (index = 0, offset = 0; index < nozzleBmdData.getEmitterCount(this->mCurrentNozzle); index += 1, offset += 0xc) {
+            MtxPtr p1 = this->getEmitMtx(index);
+            if (p1 != nullptr) {
+                *(float *)((int)&this->mEmitPos[0].x + offset) = (p1)[0][3];
+                *(float *)((int)&this->mEmitPos[0].y + offset) = (p1)[1][3];
+                *(float *)((int)&this->mEmitPos[0].z + offset) = (p1)[2][3];
```
Mrkol: This doesn't look like something they would've written. I think it should be
```
mEmitPos[index].x = p1[0][3];
mEmitPos[index].y = p1[1][3];
mEmitPos[index].z = p1[2][3];
```
which might also fix the regswap

TheAzack9: I don't disagree that this is 99% wrong, but it is the closest i could get. Doing it like suggested was further off than a register diff. Not sure exactly why, but i'm suspecting that this is some kind of inline setter as well. I will add a comment about it!

include/Player/WaterGun.hpp
TheAzack9: TIL! I will fix👍

include/Player/WaterGun.hpp
Mrkol: I think it's better to leave it matching less but making more sense. I've seen analogous problems in a couple of other places, so once we have enough examples, it should be easier to figure out the real inline (which is probably inside of TVec<f32>). But I'm not insisting, it's okay to leave as-is too =)

TheAzack9:
I'm gonna close this PR. I am gonna put up a new one when i get further :) Still working on this tho!

PR #50, opened 2025-03-03
JAudio headers

TheAzack9:
Adds headers for JAudio and empty implementations


TheAzack9:
This PR is missing JALModSe. It contains a lot of template functions, so i think it is easier to do with a more concrete use case.

PR #52, opened 2025-03-19
Implement Params and ParamsInst

TheAzack9:
Also updates MarioInit to use new param macro + initialize with proper values.

I could not make TParams::load match 100%, i could not figure out how to make it evaluate to mr r4, r31 instead of addi r4, r31, 0x0. I think the operations are basically equivalent though.
I also couldn't quite figure out how to create a named field in .rodata... It currently is evaluated to @49.

### DIFF
--- a/include/Player/MarioInit.hpp
+++ b/include/Player/MarioInit.hpp
@@ -2,471 +2,946 @@
 #define MARIOINIT_HPP
 
 #include "System/ParamInst.hpp"
-
-// TODO: move to a better file
-#include "System/BaseParam.hpp"
-class TParams {
-public:
-	char* filename;
-	TBaseParam* head;
-};
+#include "System/Params.hpp"
 
 // length 0x42C
 class TDeParams : public TParams {
 public:
-	TParamT<s16> hpMax;
-	TParamT<f32> runningMax;
-	TParamT<f32> dashMax;
-	TParamT<f32> dashAcc;
-	TParamT<f32> dashBrake;
-	TParamT<s16> dashStartTime;
-	TParamT<s16> waitingRotSp;
-	TParamT<s16> runningRotSpMin;
-	TParamT<s16> runningRotSpMax;
-	TParamT<s16> rocketRotSp;
-	TParamT<s16> pumpingRotSpMin;
-	TParamT<s16> pumpingRotSpMax;
-	TParamT<s16> invincibleTime;
-	TParamT<s16> footPrintTimerMax;
-	TParamT<u8> waterTriggerRate;
-	TParamT<u8> graffitoNoDmgTime;
-	TParamT<u8> restMax;
-	TParamT<f32> shadowSize;
-	TParamT<f32> shadowErase;
-	TParamT<f32> holdRadius;
-	TParamT<f32> damageRadius;
-	TParamT<f32> damageHeight;
-	TParamT<f32> attackHeight;
-	TParamT<f32> trampleRadius;
-	TParamT<f32> pushupRadius;
-	TParamT<f32> pushupHeight;
-	TParamT<f32> hipdropRadius;
-	TParamT<f32> quakeRadius;
-	TParamT<f32> quakeHeight;
-	TParamT<f32> tramplePowStep1;
-	TParamT<f32> tramplePowStep2;
-	TParamT<f32> tramplePowStep3;
-	TParamT<f32> jumpWallHeight;
-	TParamT<f32> throwPower;
-	TParamT<f32> slipStart;
-	TParamT<f32> wasOnWaterSlip;
-	TParamT<f32> inWaterSlip;
-	TParamT<f32> toroccoRotSp;
-	TParamT<s16> recoverTimer;
-	TParamT<s16> hotTimer;
-	TParamT<f32> feelDeep;
-	TParamT<f32> damageFallHeight;
-	TParamT<f32> forceSlipAngle;
-	TParamT<f32> clashSpeed;
-	TParamT<f32> hangWallMovableAngle;
-	TParamT<f32> colMvMax;
-	TParamT<s16> noFreezeTime;
-	TParamT<s16> kickFreezeTime;
-	TParamT<s16> surfStartFreezeTime;
-	TParamT<f32> sleepingCheckDist;
-	TParamT<f32> sleepingCheckHeight;
-	TParamT<s16> illegalPlaneCtInc;
-	TParamT<s16> illegalPlaneTime;
+	TDeParams(const char* prm)
+	    : TParams(prm)
+	    , PARAM_INIT(mHpMax, 8)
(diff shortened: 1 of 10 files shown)


include/System/BaseParam.hpp
```diff
@@ -3,12 +3,15 @@
 
 #include <JSystem/JSupport/JSUMemoryInputStream.hpp>
 
+class TParams;
+
 class TBaseParam {
 public:
-	virtual void load(JSUMemoryInputStream&);
+	TBaseParam(TParams* params, unsigned short code, const char* paramName);
+	virtual inline void load(JSUMemoryInputStream&);
```
Mrkol: Na, this is wrong, just define the function here, inside of the class. When you do non-out-of-line method definitions, the method automatically gets inline linkage and so the symbol becomes weak. But as this symbol cannot be completely optimized out as it needs to be referenced by the vtable, it will get emitted into the first TU that has a definition for any other method of the class, which in this case is precisely BaseParam.cpp

include/System/ParamInst.hpp
```diff
@@ -1,13 +1,35 @@
 #ifndef PARAM_INST_HPP
 #define PARAM_INST_HPP
 
+#include "JSystem/JDrama/JDRNameRef.hpp"
 #include "System/BaseParam.hpp"
```
Mrkol: For most of the repo I've used the following convention, so lets try to continue it to make stuff consistent:

- `<stuff.h>` is for absolute paths (relative to the `include` folder)
- `"stuff.h"` is for relative paths (relative to the current .cpp file)

And only the `<stuff.h>` is allowed to be used in headers which are inside of `include`, as relative paths would break transitive includes

include/System/ParamInst.hpp
```diff
@@ -1,13 +1,35 @@
...
 };
 
+template <typename T> class TParamRT : public TParamT<T> {
+public:
+	TParamRT(TParams* parent, u16 keycode, const char* name, T defaultValue)
+	    : TParamT<T>(parent, keycode, name, defaultValue)
+	{
+	}
+
+	inline void set(T param) {};
```
Mrkol, line 34: why?

include/System/ParamInst.hpp
```diff
@@ -1,13 +1,35 @@
...
 public:
+	TParamT(TParams* params, unsigned short code, const char* paramName,
+	        T defaultValue)
+	    : TBaseParam(params, code, paramName)
+	    , value(defaultValue)
+	{
+	}
+
 	void load(JSUMemoryInputStream& stream);
+	T get() const { return value; }
```
Mrkol, line 22: please try to mark fabricated inlines with //fabricated for future reference

include/System/ParamInst.hpp
```diff
@@ -1,13 +1,35 @@
...
 template <typename T> class TParamT : public TBaseParam {
 public:
+	TParamT(TParams* params, unsigned short code, const char* paramName,
+	        T defaultValue)
+	    : TBaseParam(params, code, paramName)
+	    , value(defaultValue)
+	{
+	}
+
 	void load(JSUMemoryInputStream& stream);
```
Mrkol: please mark virtual functions as virtual even when overriding for clarity (I'm against it when writing new code, but for decomp it's quite important I think)

src/System/Params.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+	mArc      = JKRFileLoader::getVolume("params");
+	mSceneArc = JKRFileLoader::getVolume("scene");
+}
+
+void TParams::finalize()
+{
+	mArc      = nullptr;
+	mSceneArc = nullptr;
+}
```
Mrkol: Please sort the functions as they are shown in objdiff! It can influence the order of stuff in the `data` section and make individual sections not match!

src/System/Params.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+bool TParams::load(const char* filename)
+{
+	bool found = false;
+	if (filename[0] == '/') {
+		filename++;
+	}
+
+	void* resource = nullptr;
+	if (mSceneArc != nullptr && mSceneArc->becomeCurrent("/map/params")) {
```
Mrkol: add a `static const char SceneParamsDir[] = "/map/params";` to the cpp file (don't add a declaration to the hpp, the symbol is local!) and use it here

src/System/Params.cpp
```diff
@@ -0,0 +1,74 @@
...
+		resource = mSceneArc->getResource(filename);
+	}
+
+	if (resource != nullptr) {
+		s32 size = mSceneArc->getResSize(resource);
+		JSUMemoryInputStream stream(resource, size);
```
Mrkol, line 23: removing the `(void)!buffer;` inside of the stream's constructor fixes this, but breaks J2DSetScreen::J2DSetScreen

BUT! We can place the cheeky `(void)!res;` inside of J2DSetScreen's ctor and it'll still match! make this fix pls =)
what makes me think that this is correct and the assert was actually inside of J2DSetScreen is the fact that TP debug doesn't have any asserts inside of JSUMemoryInputStream and it is extremely unlikely that they would remove such an assert in a newer jsystem version

Mrkol (changes requested):

include/Player/MarioInit.hpp
```diff
@@ -2,471 +2,946 @@
...
+	TParamT<f32> mHangWallMovableAngle;
+	TParamT<f32> mColMvMax;
+	TParamT<s16> mNoFreezeTime;
+	TParamT<s16> mKickFreezeTime;
+	TParamT<s16> mSurfStartFreezeTime;
+	TParamT<f32> mSleepingCheckDist;
+	TParamT<f32> mSleepingCheckHeight;
+	TParamT<s16> mIllegalPlaneCtInc;
+	TParamT<s16> mIllegalPlaneTime;
 };
```
Mrkol, line 121: Looking at the linker map closer, all of the classes in this header are wrong, they all have to live inside of the TMario class =/
Maybe we can do it with forward declarations, but chances are that Mariol.hpp was simply a 10K LoC long file...

TheAzack9: Fair enough, can reorganize it in another PR :) It makes sense with what Joshua did for TMario in the header interfaces.

include/System/ParamInst.hpp
TheAzack9: I had forgotten your life lesson of gw 😅

include/System/ParamInst.hpp
TheAzack9: Sounds smart, makes it easier to distinguish between known symbols and guessed ones. i assume?

src/System/Params.cpp
TheAzack9: Thanks!!

PR #53, opened 2025-04-21
WIP MSBGM Function Decompilation

ThePlayerRolo:
A WIP JAISound* MSBgm::getHandle Decompilation.
Currently 17.27% matching.
I'm new to this whole thing so any help would be appreciated!
Decomp.Me Scratch:
https://decomp.me/scratch/ghyAc

### DIFF
--- a/include/MSound/MSoundBGM.hpp
+++ b/include/MSound/MSoundBGM.hpp
@@ -8,7 +8,6 @@
 #include <JSystem/JAudio/JASystem/JASTrack.hpp>
 
 template <typename T, typename U> class JALListVirtualNode;
-
 class MSBgm : public JALListHioNode<MSBgm, u32> {
 public:
 	MSBgm();
@@ -33,6 +32,14 @@ public:
 	static u32 getSceneNo(u32);
 	static JAISound* getHandle(u8);
 	static JAISound* getJASTrack(JAISound*, u8);
+	//TODO: Found out where this inline is from.
+	static JAISound* someInline() {
+        if (smBgmInTrack[0] != 0) {
+    		return smBgmInTrack[0]->unk_14;
+    	} else {
+            return nullptr;
+        }
+    }
 public:
 	/* 0x14 */ JAISound* unk_14;
 };
--- a/include/MSound/MSoundBGM.hpp
+++ b/include/MSound/MSoundBGM.hpp
@@ -35,13 +35,13 @@ public:
 	//TODO: Found out where this inline is from.
 	static JAISound* someInline() {
         if (smBgmInTrack[0] != 0) {
-    		return smBgmInTrack[0]->unk_14;
+    		return smBgmInTrack[0]->unk14;
     	} else {
             return nullptr;
         }
     }
 public:
-	/* 0x14 */ JAISound* unk_14;
+	/* 0x14 */ JAISound* unk14;
 };
 
 #endif
--- a/include/MSound/MSoundBGM.hpp
+++ b/include/MSound/MSoundBGM.hpp
@@ -14,8 +14,8 @@ public:
 	MSBgm();
 
 	static JASystem::TTrack* smBgmInTrack[3];
-	static f32 smMainVolume;
-
+	static MSBgm* smMainVolume;
+	//static f32 smMainVolume;
 	static void init();
 	static void startBGM(u32);
 	static void stopBGM(u32, u32);
--- a/include/MSound/MSoundBGM.hpp
+++ b/include/MSound/MSoundBGM.hpp
@@ -13,9 +13,8 @@ class MSBgm : public JALListHioNode<MSBgm, u32> {
 public:
 	MSBgm();
 
-	static JASystem::TTrack* smBgmInTrack[3];
-	static MSBgm* smMainVolume;
-	//static f32 smMainVolume;
+	static MSBgm* smBgmInTrack[3];
+	static f32 smMainVolume;
 	static void init();
 	static void startBGM(u32);
 	static void stopBGM(u32, u32);
--- a/include/MSound/MSoundBGM.hpp
+++ b/include/MSound/MSoundBGM.hpp
@@ -12,7 +12,6 @@ template <typename T, typename U> class JALListVirtualNode;
 class MSBgm : public JALListHioNode<MSBgm, u32> {
 public:
 	MSBgm();
-	JAISound* unk_14; //0x14 Offset  
 	static MSBgm* smBgmInTrack[3];
 	static f32 smMainVolume;
 	static void init();
@@ -34,6 +33,8 @@ public:
 	static u32 getSceneNo(u32);
 	static JAISound* getHandle(u8);
 	static JAISound* getJASTrack(JAISound*, u8);
+public:
+	/* 0x14 */ JAISound* unk_14;
 };
 
 #endif
--- a/include/MSound/MSoundBGM.hpp
+++ b/include/MSound/MSoundBGM.hpp
@@ -32,14 +32,16 @@ public:
 	static u32 getSceneNo(u32);
 	static JAISound* getHandle(u8);
 	static JAISound* getJASTrack(JAISound*, u8);
-	//TODO: Found out where this inline is from.
-	static JAISound* someInline() {
-        if (smBgmInTrack[0] != 0) {
-    		return smBgmInTrack[0]->unk14;
-    	} else {
-            return nullptr;
-        }
-    }
+	// TODO: Found out where this inline is from.
+	static JAISound* someInline()
+	{
+		if (smBgmInTrack[0] != 0) {
+			return smBgmInTrack[0]->unk14;
+		} else {
+			return nullptr;
+		}
+	}
+
 public:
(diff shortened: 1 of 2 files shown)


break-core:
This is a pretty good first PR.

I cleaned this up a bit and got it to a much higher percentage (i'm also a noob at this too, dw). Only one instruction left to do
https://decomp.me/scratch/ET5lj

The `!= 0` is not technically needed in this case as not having it there compiles the same as having it there. Making `smBgmInTrack[param]` a temporary variable also helps the score a bit too I think? (though, this may not always be the case)

Mrkol:
You will need to reverse some structure layouts in ghidra to properly match this, request access to the ghidra repo and I'll approve it.

ThePlayerRolo:
I already got permission, idk about BreakCore though.

ThePlayerRolo:
Added a 96% matching Decompilied GetSceneNo Function.

ThePlayerRolo:
GetSceneNo is now 100% matching (LETS GOOOOOOOOOOOOOOOOOOO)

ThePlayerRolo:
Though, in the future we may have to edit it to take predefined level Varaibles instead of direct  registers (0x0000) for the returns.

ThePlayerRolo:
A very WIP Version of SetTrack that doesn't currently build.
Scratch: https://decomp.me/scratch/pErP2

ThePlayerRolo:
Currently it still doesn't build however it is now when JAISound:: is added to set volume. however removing makes it work perfectly.
Currently its around 74% matching

src/MSound/MSoundBGM.cpp
```diff
@@ -19,7 +19,17 @@ void MSBgm::setVolume(u32 param1, f32 param2, u32 param3, u8 param4)
 	MSBgm* bgm = JALListS<MSBgm, u32>::search(param1 & 0x3FF);
 }
 
-void MSBgm::setTrackVolume(u8 param1, f32 param2, u32 param3, u8 param4) { }
+void MSBgm::setTrackVolume(u8 param1, f32 param2, u32 param3, u8 param4) {
+	JAISound* sound = nullptr;
+    
+    JASystem::TTrack* track = smBgmInTrack[param1];
+	if ((track) &&
+	   (sound = *(JAISound **)((&smBgmInTrack)[param1] + 0x14), sound != (JAISound *)0x0)) {
```
Mrkol: I'm pretty sure that Azack guessed the type for smBgmInTrack incorrectly, hence why this looks like such a mess. We need to investigate what type it really is =\

Mrkol: Yes, indeed. I think I've figured it out, the real type is `MSBgm* smBgmInTrack[3]`. Basically, if you use the "search for references" ghidra thing on  `smMainVolume`, you will quickly find that it's initialized inside of `MSBgm::startBGM`, but it is initialized with the result of `JALListS<T>::search`. If my previous analysis of these awful templates is to be trusted, this method returns a `T*`, which in this case is a `MSBgm*`.
@ThePlayerRolo

ThePlayerRolo: interesting, how would you fix the error?

Mrkol: first, you fix the type of the global I mentioned above. then you figure out the structure's layout in ghidra (there's only a single field except for the parent it seems, so no real need to use ghidra). then you fix the code. I'm not sure what error you are talking about

ThePlayerRolo:
I changed smMainVolume to a BGM with no visible difference, IDK if thats what you mean as its kind of confusing. BTW, the error occurs when trying to call JAISound:SetVolume instead of BGM:SetVolume. THe asm to C code in Ghidra says JAISOund:SetVolume however I get a static type error whenever I try to do it.  THis Decomp.me displays this: https://decomp.me/scratch/pErP2

include/MSound/MSoundBGM.hpp
```diff
@@ -14,8 +14,8 @@ class MSBgm : public JALListHioNode<MSBgm, u32> {
 	MSBgm();
 
 	static JASystem::TTrack* smBgmInTrack[3];
-	static f32 smMainVolume;
-
+	static MSBgm* smMainVolume;
```
Mrkol: Wrong variable, MainVolume is definitely a f32. `smBgmInTrack` is the variable which has the wrong type, which makes the rest of the code go haywire. It should be an array of MSBgm pointers rather than JASystem::TTrack ones

ThePlayerRolo: OHHHHHHHHHHHHHH Mybad

ThePlayerRolo:
Ok so I changed smBgmInTrack  to  `MSBgm* MSBgm::smBgmInTrack[3];`   however the JAISound error still occurs.

ThePlayerRolo:
Until more is decompiled, this is the best I could do.

src/MSound/MSoundBGM.cpp
```diff
@@ -19,7 +19,22 @@ void MSBgm::setVolume(u32 param1, f32 param2, u32 param3, u8 param4)
...
 }
 
-void MSBgm::setTrackVolume(u8 param1, f32 param2, u32 param3, u8 param4) { }
+void MSBgm::setTrackVolume(u8 param1, f32 param2, u32 param3, u8 param4)
+{
+	JAISound* sound = nullptr;
+
+	MSBgm* track = smBgmInTrack[param1];
+	if ((track)
+	    && (sound = *(JAISound**)(track + 0x14), sound != (JAISound*)0x0)) {
```
Mrkol: when you see `*(T*)(x + 0x69)`, it means there's a field of type `T` in `x`'s type. Please add the fields and use them rather than dirty offset hacks and casts. Furthermore, half of the work when decomping stuff is figuring out what fields are there in each struct. To do this, you use ghidra's struct editor, look through code that uses the type and try to find places where something like `*(T*)(x + 0x69)` happens and add a field at this offset in the struct editor -- voila, ghidra's decompilation changes to use the actual field. Do this a whole bunch and then you have the entire struct/class figured out

src/MSound/MSoundBGM.cpp
```diff
@@ -19,7 +19,22 @@ void MSBgm::setVolume(u32 param1, f32 param2, u32 param3, u8 param4)
...
+		// Using JAISound::setVolume  instead causes an  Static Error. Even
+		// though that is in the Decompiled Ghidra Code. Function should be
+		// something like JAISound::setVolume(param2,param3,param4); But due to
+		// a "a nonstatic member reference must be relative to a specific
+		// object" error with JAISound I can't
```
Mrkol: Ghidra doesn't know what methods are, so it shows all functions as just that -- functions. But a method can be static too, which would make it a regular function and not really a method. In this particular case, we've already figured out that JAISound::setVolume is a regular method and not a static one, so to call it you need an object on which you will be invoking the method, something like `sound->setVolume(...)`

src/MSound/MSoundBGM.cpp
Mrkol: Furthermore, because devs of ghidra are deranged individuals, it's decompiler likes to generate constructions like these: assignment stuffed in the middle of an "if" condition, chained with the actual condition via the comma operator. Real humans don't write code like this: this isn't something the original dev's could've ever written and it isn't something that will be convenient to modify for modders.

Instead of this awfulness, when you add the proper MSBgm field, you should simply use it directly (`track->unk14`) and get rid of the `sound` variable.

Mrkol (changes requested):

src/MSound/MSoundBGM.cpp
ThePlayerRolo: I figured the + 0x14 when working on some of MarioAccess.cpp. GOing to work on that

src/MSound/MSoundBGM.cpp
ThePlayerRolo: Ah, makes sense, I was kind of messing with pointer stuff, I'm still new to pointers haha

ThePlayerRolo:
0x14 offset variable added (unk_14), getHandle, SetTrackVolume, SetDolby, SetPan, and SetVolume all 100% matching

ThePlayerRolo:
Most functions are matching. StartBGM, setYoshiPercussion, and stopTrackBgm are mostly matching, with the main diffs being inline related or ones thats will (mostly likely) fix themselves with more decompilation. I tried to decompile Init, but its just too confusing for me, so I think this is good enough for now for merging

include/MSound/MSoundBGM.hpp
```diff
@@ -35,6 +32,16 @@ class MSBgm : public JALListHioNode<MSBgm, u32> {
...
+	//TODO: Found out where this inline is from.
+	static JAISound* someInline() {
+        if (smBgmInTrack[0] != 0) {
+    		return smBgmInTrack[0]->unk_14;
+    	} else {
+            return nullptr;
+        }
+    }
+public:
+	/* 0x14 */ JAISound* unk_14;
```
Mrkol: rename to unk14 please for consistency
also clang-format the code and we can merge

Mrkol (changes requested):

PR #54, opened 2025-04-21
Initial push on MarioGamePad

TheAzack9:
This is probably far from the hard truth, but was the closest i could get at current skill level.


include/Camera/Camera.hpp
```diff
@@ -129,7 +129,7 @@ class CPolarSubCamera : JDrama::TLookAtCamera {
 	s16 getUnk2C8() const { return unk2C8; }
 
 public:
-	s32 unk50;
+	int mMode;
 	char filler1[0x2C8 - 0x54];
 	s16 unk2C8;
```
Mrkol, line 134: please add `/* 0x50 */` etc comments before fields whenever you see them missing, they are very useful to quickly look up the field you want while decomping

include/JSystem/JUtility/JUTGamePad.hpp
```diff
@@ -238,7 +261,7 @@ struct JUTGamePad : public JKRDisposer {
 	static u8 mPadAssign[4];
 	static u32 mSuppressPadReset;
 	static s32 sAnalogMode;
-	static u32 mRumbleSupported;
+	static u32 sRumbleSupported;
```
Mrkol: I think this is a rogue field left here from copy-pasting the code from tww or something, it's not present in our linker map so should be removed

include/System/MarioGamePad.hpp
```diff
@@ -1,20 +1,306 @@
...
+	s16 mStickVS16;
+	Buttons mInput;
+	Buttons mFrameInput;
+	u8 mAnalogRU8;
+	u8 mAnalogLU8;
+	f32 mStickH;
+	f32 mStickV;
+	f32 mStickDist;
+	f32 mAnalogL;
+	f32 mAnalogR;
```
Mrkol: offsets comments pls

src/System/MarioGamePad.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+	if (this->mDisabledFrames <= 0) {
+		bool isLButtonCameraSpecifyMode
+		    = gpCamera->isLButtonCameraSpecifyMode(gpCamera->mMode);
+		if (isLButtonCameraSpecifyMode) {
+			updateMeaningLCamera(prevMeaning);
+		} else {
+			updateMeaningRegular2(prevMeaning);
+		}
+	}
```
Mrkol: Maybe all of this is yet another inline and it's early-returning rather than gotoing?

include/System/MarioGamePad.hpp
```diff
@@ -1,20 +1,306 @@
...
+		this->updateMeaning(L, MEANING_0x2000, prevMeaning);
+	}
+	// Fabricated
+	inline void updateMeaning0x80(u32 prevMeaning)
+	{
+		this->updateMeaning(A, MEANING_0x20, prevMeaning);
+		this->updateMeaning(B, MEANING_0x40, prevMeaning);
+		this->mCompSPos[2].x = this->mMainStick.mPosX;
+		this->mCompSPos[2].y = this->mMainStick.mPosY;
+	}
```
Mrkol: how did you decide on these particular inlines? does it help matching it except for stack size?
because it doesn't look typical to me for them to put so much code into headers, tbh =\
I can understand `updateMeaning(EButtons button, PadMeanings padButton, u32 prevMeaning)`, but the others look way too unrealistic

Mrkol (changes requested):
fix the build pls

src/System/MarioGamePad.cpp
TheAzack9: I tried and retried now to make sure, but it then decides to just inline the code at each if part instead of creating jumps to the end in the asm.
I still suspect it to be a switch statement of sorts, but i could never make that match.

include/System/MarioGamePad.hpp
TheAzack9: They are most likely wrong.
The reason i split them is because based on which "mode" the controller is in it behaves differently. I abstracted each "mode" into a separate inline function because i feel the very large size was hinting towards that. I didn't spend enough time to figure out exactly what each mode is, but some are guis, some are different in game states like y-cam and regular.
iirc, doing that split didn't even affect stack size, which makes it even more likely that it is wrong... I can flatten it if wanted tho!

include/System/MarioGamePad.hpp
```diff
@@ -1,20 +1,306 @@
...
+		this->mCompSPos[1].y = (f32)this->mButton.mAnalogR;
+
+		this->updateMeaning(R, MEANING_0x400, prevMeaning);
+		this->updateMeaning(Z, MEANING_0x1000, prevMeaning);
+		this->updateMeaning(L, MEANING_0x2000, prevMeaning);
+		this->mCompSPos[2].x = this->mMainStick.mPosX;
+		this->mCompSPos[2].y = this->mMainStick.mPosY;
+		this->updateMeaning(A, MEANING_0x10000, prevMeaning);
+		this->updateMeaning(B, MEANING_0x10000, prevMeaning);
+		this->updateMeaning(Y, MEANING_0x4000, prevMeaning);
```
Mrkol: I haven't used `this->` prefix in most of the code, so I think we should keep following that convention. I'll write codestyle guidelines at some point to fix stuff like this in stone

include/System/MarioGamePad.hpp
Mrkol: After giving it some thought, yes, lets flatten it back, as that makes a lot more sense in the context of how the rest of the game was written (except for the `updateMeaning(EButtons button, PadMeanings padButton, u32 prevMeaning)` inline which makes a lot of sense)

Mrkol (changes requested):

PR #55, opened 2025-05-13
Start GC2D/hx_wiper.c - match some functions, add definitions/empty functions

LMFinish:
Matrix code may need review but is otherwise matched


Mrkol:
Seems to be abandoned -- closing.

PR #59, opened 2025-08-13
Beam mostly matching

Cuyler36:
calcVertices has problems


include/JSystem/JGeometry/JGVec3.hpp
```diff
@@ -141,6 +146,12 @@ template <> class TVec3<f32> : public Vec {
 		return fst;
 	}
 
+	friend TVec3 operator*(TVec3 fst, float snd)
```
Mrkol, line 151: Add fabricated comments for all new fucs here pls

PR #60, opened 2025-09-05
Mostly decompile __num2dec

hyblocker:
mostly decompiles `__num2dec` from `ansi_fp.c`. final asm needs a bit of work to hit 100% but at least its mostly there!


hyblocker:
implements:
```
inverse_trig.c (100%)
ansi_fp.c (100%)
```
makes `trigf.c` match 100% too

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/inverse_trig.c
```diff
@@ -0,0 +1,164 @@
+#include "PowerPC_EABI_Support/Msl/MSL_C/MSL_Common/math.h"
+#include "dolphin/types.h"
+
+#define __PI_O2 ((f32)M_PI / 2)
```
Mrkol: Can you quickly get rid of all dolphin types, please? Original couldn't have used them because libc is by metrowerks and is cross platform so couldn't have depended on dolphin SDK headers

Mrkol (changes requested):

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/inverse_trig.c
```diff
@@ -0,0 +1,164 @@
...
+ * Size:	000020
+ */
+__declspec(weak) f32 atan__Ff(f32 x) { return atanf(x); }
+
+#pragma dont_inline reset
+/*
+ * --INFO--
+ * Address:	8021B730
+ * Size:	0001F4
+ */
```
Mrkol: These address comments are from a different v repo and different binary so let's get rid of them

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/inverse_trig.c
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return HUGE_VALF;
+}
+
+/*
+ * --INFO--
+ * Address:	8021B924
+ * Size:	000020
+ */
+__declspec(weak) f32 atan__Ff(f32 x) { return atanf(x); }
```
Mrkol: Nope, this should be simply marked as inline instead of weak, defined under pragma cplusplus on and put into one of the headers like math.h

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/trigf.c
```diff
@@ -153,8 +147,8 @@ float cosf(float x)
 	                                    // n*pi/4 is a multiple of pi/2(not pi)
 }
 
-__declspec(weak) float cos__Ff(float x) { return cosf(x); }
 __declspec(weak) float sin__Ff(float x) { return sinf(x); }
+__declspec(weak) float cos__Ff(float x) { return cosf(x); }
```
Mrkol, line 151: Same for these, put into a header, wrap with cplusplus on, inline instead of declspec weak. It probably didn't inline it because inlining cpp code into c is not allowed rather than because of pragma dont_inline. Oh, and remove manual mangling

Mrkol (changes requested):
Also mark files that match 100% as Matching in the configure.py script please

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/inverse_trig.c
hyblocker: If I move it to the header it just disappears from the TU altogether, and `acosf` ends up deciding to invoke `atanf` instead of `atan(float)`. also order of the functions gets lost

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/trigf.c
hyblocker: if i remove move them to the header they just disappear from the TU altogether and actually get inlined. `tanf` ends up invoking `cosf` and `sinf` directly. ditto with function ordering

src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/inverse_trig.c
Mrkol: Ok, let's merge as is then, I'll try to fix it too afterwards

PR #61, opened 2025-09-25
Decompile THPRender

hyblocker:
Gets a 100% match, however `JDrama::TViewObj`'s vtable and destructor end up getting defined in this TU, polluting the `.data-0` section (thus also fails to match as a result). Otherwise it's a 100% match.

### DIFF
--- a/include/System/THPRender.hpp
+++ b/include/System/THPRender.hpp
@@ -1,0 +1,24 @@
+#ifndef SYSTEM_THP_RENDER_HPP
+#define SYSTEM_THP_RENDER_HPP
+
+#include <JSystem/JDrama/JDRViewObj.hpp>
+#include <JSystem/JDrama/JDRGraphics.hpp>
+#include "dolphin/gx/GXTransform.h"
+
+#include "MarioUtil/DrawUtil.hpp"
+#include "THPPlayer/THPPlayer.h"
+
+class TTHPRender : public JDrama::TViewObj {
+public:
+	virtual ~TTHPRender() {}
+	void perform(u32, JDrama::TGraphics*); /* override */
+	TTHPRender(const char* name);
+
+	/* 0x10 */ u32 x;
+	/* 0x14 */ u32 y;
+	/* 0x18 */ u32 polyW;
+	/* 0x1c */ u32 polyH;
+	/* 0x20 */ u32 unk10;
+};
+
+#endif // SYSTEM_THP_RENDER_HPP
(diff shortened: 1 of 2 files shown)


src/System/THPRender.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+void TTHPRender::perform(u32 flags, JDrama::TGraphics* gfx)
+{
+	if ((flags & 8)) {
+
+		SMS_DrawInit();
+		GXLoadPosMtxImm(gfx->unkB4, GX_PNMTX0);
+		GXSetCurrentMtx(GX_PNMTX0);
+		this->frameNumber = THPPlayerDrawCurrentFrame(0, this->x, this->y,
+		                                              this->polyW, this->polyH);
```
Mrkol: No need to use "this" here

src/System/THPRender.cpp
```diff
@@ -0,0 +1,20 @@
...
+	}
+}
+TTHPRender::TTHPRender(const char* name)
```
Mrkol, line 14: leave  a newline between functions and at the end of file, let 'em breathe

include/System/THPRender.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef SYSTEM_THP_RENDER_HPP
+#define SYSTEM_THP_RENDER_HPP
+
+#include <JSystem/JDrama/JDRViewObj.hpp>
+#include <JSystem/JDrama/JDRGraphics.hpp>
+#include "dolphin/gx/GXTransform.h"
+
+#include "MarioUtil/DrawUtil.hpp"
+#include "THPPlayer/THPPlayer.h"
```
Mrkol: I've been using `<>` brackets almost everywhere for absolute include paths, so better to follow the convention

include/System/THPRender.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+class TTHPRender : public JDrama::TViewObj {
+public:
+	virtual ~TTHPRender() { }
```
Mrkol: you don't need to manually define the destructor, it gets automatically defined

include/System/THPRender.hpp
```diff
@@ -0,0 +1,24 @@
...
+	void perform(u32, JDrama::TGraphics*); /* override */
```
Mrkol: lets mark all `virtual` functions as such (cuz we don't have override)

Mrkol (changes requested):

PR #63, opened 2025-10-12
M3UModel - changeMtxCalcAnmTransform 100% objdiff, unk4 and unk14 types

MattWBowen:
used changeAnmTexPattern as a base, eventually got this, and it gave 100% objdiff match


Mrkol:
Hi, sorry, but for now I don't see enough proof that you've guessed unk14 and unk4's type correctly. Especially that "offset by param_1 * 6" thing is suspicious, it probably was a struct at the very least, or maybe an array. For this to be merged you will need to find and tackle more functions that use these fields so that we know the guess is correct for sure, preferably find the functions that initialize these fields.

PR #65, opened 2025-10-19
Decompile TBathWaterManager::draw_mist

TheAzack9:
99% match, one wrong register + data due to other functions.

### DIFF
--- a/include/Map/BathWaterManager.hpp
+++ b/include/Map/BathWaterManager.hpp
@@ -4,6 +4,8 @@
 #include <Strategic/HitActor.hpp>
 #include <JSystem/JMath.hpp>
 
+static void draw_mist(u16 x, u16 y, u16 wd, u16 ht, void* buffer);
+
 class TBathtubData {
 public:
 	void getGravityDir(f32) const;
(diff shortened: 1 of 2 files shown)


include/Map/BathWaterManager.hpp
```diff
@@ -4,6 +4,8 @@
 #include <Strategic/HitActor.hpp>
 #include <JSystem/JMath.hpp>
 
+static void draw_mist(u16 x, u16 y, u16 wd, u16 ht, void* buffer);
```
Mrkol: Don't declare static functions in headers, static functions are by definition "local" to the cpp file they are defined inside of and cannot be called from any other cpp file

Mrkol (changes requested):

include/Map/BathWaterManager.hpp
TheAzack9: Fair, had a weird brain fart here thinking i had to do it like this to make it formatted correctly. Saw now that i never marked it as static in the cpp. 😅  Fixed now 😄

PR #66, opened 2025-10-20
Add TBathtub and add stubs for functions

TheAzack9:
I am unsure what conventions are here, for me this feels like TBathtub could've been in a separate file. However from symbols it hints that it should be in MapObjCorona. 🤔 Been a while so i forgot, so please yell if i should move it.

### DIFF
--- a/src/MoveBG/MapObjCorona.cpp
+++ b/src/MoveBG/MapObjCorona.cpp
@@ -3,13 +3,13 @@
 
 void TBathtub::loadAfter() { }
 
-void TBathtub::hipdrop(const JGeometry::TVec3<float>&) { }
+void TBathtub::hipdrop(const JGeometry::TVec3<f32>&) { }
 
-void TBathtub::quake(const JGeometry::TVec3<float>&) { }
+void TBathtub::quake(const JGeometry::TVec3<f32>&) { }
 
 u8 TBathtub::getNumGripsDead() const { return 0; }
 
-void TBathtub::tumble(float, float) { }
+void TBathtub::tumble(f32, f32) { }
 
 MtxPtr TBathtub::getTakingMtx() { return nullptr; }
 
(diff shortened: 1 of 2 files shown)


src/MoveBG/MapObjCorona.cpp
```diff
@@ -0,0 +1,104 @@
...
+// Unused
+MtxPtr TBathtub::getShineMtx() { return nullptr; }
+
+// Unused
+void TBathtub::liftMario(const JGeometry::TVec3<f32>&) { }
+
+// Unused
+void TBathtub::trample(const JGeometry::TVec3<f32>&) { }
+
+TBathtub::~TBathtub() { }
```
Mrkol: Remove the ctor definition & declaration,  I'm 99.99% sure it's auto-generated and the symbol is weak. Check other symbols in this TU too please, in case they are weak and should be defined in the header rather than here

Mrkol (changes requested):

src/MoveBG/MapObjCorona.cpp
TheAzack9: I assume you meant dtor? You were at least 100% correct there :)

I checked the other symbols, and at least everything that is used does not seem to be weak. I have no idea with the unused symbols ofc.

TheAzack9:
This is a bit old, and i think i would disagree with myself now on placement of the symbols. Would it be better that i moved the header to a Bathtub.hpp and keep the stubs in MapObjCorona.cpp?

Mrkol:
> This is a bit old, and i think i would disagree with myself now on placement of the symbols. Would it be better that i moved the header to a Bathtub.hpp and keep the stubs in MapObjCorona.cpp?

Nah, I think this is fine =)

src/MoveBG/MapObjCorona.cpp
TheAzack9: Removed the destructor now :)

PR #67, opened 2025-12-28
Fix command to unquarantine Wine Crossover.app

benny-dreamly:
the -r flag in the `xattr` command is not always supported, and inadvisable in a template command


Mrkol:
DTK-template needs to be updated to upstream, after which the need to use wine-crossover will disappear.

PR #70, opened 2026-01-30
100% TMario::actnMain

TheAzack9:
Starting my new years resolution of not making huge PRs :)

Edit:
Maybe for next year it should be to proof read my code before making a PR...

### DIFF
--- a/include/Player/MarioMain.hpp
+++ b/include/Player/MarioMain.hpp
@@ -6,6 +6,9 @@
 #include <System/ParamInst.hpp>
 #include <System/DrawSyncCallback.hpp>
 #include <Strategic/TakeActor.hpp>
+#include <JSystem/J3D/J3DGraphBase/J3DDrawBuffer.hpp>
+#include <JSystem/J3D/J3DGraphAnimator/J3DModel.hpp>
+#include <M3DUtil/M3UModelMario.hpp>
 
 class TLiveActor;
 class TBGCheckData;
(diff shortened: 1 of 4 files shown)


include/Player/MarioMain.hpp
```diff
@@ -637,7 +640,7 @@ class TMario : public TTakeActor, public TDrawSyncCallback {
 	void changeHandByRate(f32);
 	void changeHand(int);
 	void isAnimeLoopOrStop();
-	void isLast1AnimeFrame();
+	s32 isLast1AnimeFrame();
```
Mrkol: `BOOL` instead of s32 if `bool` doesn't match

include/Player/MarioMain.hpp
```diff
@@ -779,7 +782,7 @@ class TMario : public TTakeActor, public TDrawSyncCallback {
 	void changePos(const Vec&);
 	void isSpeedZero();
 	void canBendBody();
-	void considerRotateJumpStart();
+	s32 considerRotateJumpStart();
```
Mrkol: Same here, if it's a bool-ish return and bool doesn't match -- use BOOL

Mrkol (changes requested):

include/Player/MarioMain.hpp
TheAzack9: Fixed both of these :)

PR #71, opened 2026-02-01
match TMario::taking

TheAzack9:


TheAzack9:
Should i put the file as matching now, or should it wait until __sinit_MarioAction_cpp is matching as well?

PR #75, opened 2026-02-03
TCoasterKiller 92%

Cuyler36:
Gets coaster to ~92%. Most of the remaining work needs to be done in JGeometry::TQuat4<T>'s methods. I referenced what was in the SMG1 repo to scaffold the functions to use.

Also includes some work on TBathtubKiller and TRocket that were necessary or helpful.


include/Enemy/CoasterKiller.hpp
```diff
@@ -0,0 +1,113 @@
...
+	virtual void perform(u32, JDrama::TGraphics*);
+	virtual void reset();
+	
+	TCoasterEnemyParams* getSaveParam2() const { return (TCoasterEnemyParams*)getSaveParam(); }
+	int getPathIdx() const { return pathIdx; }
+	int getPathDir() const { return pathDir; }
+	const JGeometry::TQuat4<f32>& getQuat() const { return mQuat; }
+
+	/* 0x194 */ int pathIdx;
+	/* 0x198 */ int pathDir;
```
Mrkol: we do `m` prefixing in this neighborhood

Mrkol (changes requested):

PR #76, opened 2026-02-06
MarioAutodemo.cpp mostly matching

TheAzack9:
I copied the s16 implementation for set for TVec<s16> from TP since i couldn't make it match properly. I couldn't see the existing one being used anywhere, so i kept the old implementation as a comment until the PR.

If the existing one has been validated, then i will revert the copy and let the unmatching stuff be until we have more examples. Otherwise i will remove the old implementation for it.


src/Player/MarioAutodemo.cpp
```diff
@@ -85,3 +93,475 @@ BOOL TMario::jumpingDemoCommon(u32 playerStatus, int animationId, f32 velocity)
...
+	holderPosOffset.y -= 80.0f;
+	switch (mActionState) {
+	case 0:
+		if (mActionTimer <= 1) {
+			if (onYoshi() != FALSE) {
+				getOffYoshi(true);
+			}
+			TModelGate* gate  = (TModelGate*)mHolder;
+			u16 gateUnk72     = gate->unk72;
+			Mtx* nodeMatrices = gate->unk78->getModel()->mNodeMatrices;
```
Mrkol: ghidra is lying, you can just do `MtxPtr mtx = gate->unk78->getModel()->getAnmMtx(gateUnk72);` and use it below and it'll still match

Mrkol (changes requested):

src/Player/MarioAutodemo.cpp
TheAzack9: Ghidra tends to do that 😅  I should really get better at checking for existing better functions :) Thanks for the heads up!

PR #77, opened 2026-02-06
TMario scaffolding pt 1

Cuyler36:
Decent amount of work scaffolding TMario actor class members. Still needs work. Gave variable names based on usage unless they're unused or low ref count. In those cases I left them as unknown.

### DIFF
--- a/include/MarioUtil/MtxUtil.hpp
+++ b/include/MarioUtil/MtxUtil.hpp
@@ -2,40 +2,118 @@
 #define MTX_UTIL_HPP
 
 #include <JSystem/JGeometry.hpp>
+#include <System/Params.hpp>
+#include <System/ParamInst.hpp>
 
 class J3DNode;
 class J3DModel;
 class THitActor;
 
+enum TMtxEffectType {
+	TMTX_EFFECT_TIME_LAG,
+	TMTX_EFFECT_SWING_RZ,
+	TMTX_EFFECT_SWING_RZ_REVERSE_XZ,
+
+	TMTX_EFFECT_TYPE_MAX
+};
+
 void MtxToQuat(MtxPtr, Quaternion*);
 
 class TMtxEffectBase {
 public:
-	void getParams();
+	virtual TParams* getParams() { return nullptr; }
+
+	u16 mFlags;
 };
 
-class TMtxTimeLag {
+class TMtxTimeLag : public TMtxEffectBase {
 public:
-	void getParams();
-	struct TDeParams {
-		TDeParams(const char*);
+	class TDeParams : public TParams {
+	public:
+		// ctor exists unused in symbol map (inlined)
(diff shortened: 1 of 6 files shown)


include/MarioUtil/MtxUtil.hpp
```diff
@@ -1,41 +1,118 @@
...
+	TMTX_EFFECT_SWING_RZ_REVERSE_XZ,
+
+	TMTX_EFFECT_TYPE_MAX
+};
+
 void MtxToQuat(MtxPtr, Quaternion*);
 
 class TMtxEffectBase {
 public:
-	void getParams();
```
TheAzack9, line 14: I am wondering if the TMtxEffectBase might have some flag field. At the bottom of TMultiMtxEffect::setup it does add 2, as well as in TMarioCap::mtxEffectHide/Show i know it also switches on and off the number 1 bit. Did you not see similar things in other places it was used? 🤔
Otherwise i have probably made some invalid assumptions lmao

src/Player/MarioAutodemo.cpp
```diff
@@ -423,23 +423,22 @@ BOOL TMario::electricDamage()
...
-		unk484.x = mPosition.x + JMASSin(mFaceAngle.y);
-		unk484.z = mPosition.z + JMASCos(mFaceAngle.y);
+		mFloorHitActor.mPosition.x = mPosition.x + JMASSin(mFaceAngle.y);
+		mFloorHitActor.mPosition.z = mPosition.z + JMASCos(mFaceAngle.y);
 
 		// This matches, but i am pretty certain it is wrong
 		// damageExec((THitActor*)&unk474, 0, 3,
 		//            mDmgParamsGraffitoElec.mWaterEmit.get(),
 		//            mDmgParamsGraffitoElec.mMinSpeed.get(),
 		//            mDmgParamsGraffitoElec.mMotor.get(), 0.0f, 0x3C);
```
TheAzack9: please remove outdated comment :)

include/MarioUtil/MtxUtil.hpp
Cuyler36: Oh it does, I just missed moving it over! Thanks!

src/Player/MarioAutodemo.cpp
Cuyler36: Sure, I'll move the correct stuff in this PR

include/MarioUtil/MtxUtil.hpp
```diff
@@ -1,41 +1,120 @@
 #ifndef MTX_UTIL_HPP
 #define MTX_UTIL_HPP
 
+#include "JSystem/JGeometry/JGQuat4.hpp"
```
Mrkol: use same brackets everywhere please, i.e. <>

PR #85, opened 2026-02-20
MarioCollision.cpp mostly matching

TheAzack9:


TheAzack9:
Again, should really fix the linter. idk why, but my clang-format insists on E_SIDEWALK_TYPE not having spaces, which breaks with the pipeline

PR #87, opened 2026-02-27
Work on MarioDraw.cpp

TheAzack9:
This is more fuzzy than i wished for, but i struggled with MarioDraw lmao, so much data...

### DIFF
--- a/include/M3DUtil/M3UModelMario.hpp
+++ b/include/M3DUtil/M3UModelMario.hpp
@@ -2,7 +2,7 @@
 #define M3DUTIL_M3U_MODEL_MARIO_HPP
 
 #include <M3DUtil/M3UModel.hpp>
-// #include <JSystem/J3D/J3DGraphAnimator/J3DJoint.hpp>
+#include <JSystem/J3D/J3DGraphAnimator/J3DJoint.hpp>
 
 class M3UMtxCalcSIAnmBlendQuat;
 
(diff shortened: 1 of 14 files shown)


src/Player/MarioDraw.cpp
```diff
@@ -1,7 +1,2129 @@
...
+		                  mBodyModelData);
+		u16 matCount   = anmTexPattern[i]->getUpdateMaterialNum();
+		anmTexNoAnm[i] = new J3DTexNoAnm[matCount];
+
+		for (int j = 0; j < matCount; ++i) {
+			anmTexNoAnm[i][j].setAnmIndex(j);
+			anmTexNoAnm[i][j].setAnmTexPattern(anmTexPattern[i]);
+		}
+		i++;
+	} while (i < 0x18);
```
Mrkol: Should definitely be a regular for loop

include/M3DUtil/M3UModelMario.hpp
```diff
@@ -33,6 +39,7 @@ class M3UModelMario : public M3UModel {
 public:
 	/* 0x20 */ M3UModelCommonMario* unk20;
 	/* 0x24 */ M3UMarioMtxCalcSetInfo* unk24;
+	// /* 0x28 */ J3DMtxCalcSoftimage unk28;
```
TheAzack9: I am going to take a second look at M3UModelMario for this, but this was causing some of the broken matches bcs of the `#include <JSystem/J3D/J3DGraphAnimator/J3DJoint.hpp>`. It is not in use currently so i just commented it out, but initModel in TMario refers to this and adding this makes it match.

I assume we possibly need the stuff with precompiled headers 🤔

TheAzack9:
I think i will stop with MarioDraw.cpp for now. 😅 calcBaseMtx drained me lmao

PR #92, opened 2026-03-13
MarNameRefGen

hyblocker:
mostly decomp the MarNameRefGen classes; can't be 100% matching yet because we still need to decompile the classes they reference


Mrkol:
Why'd you put the name into the ctor's default value in some cases but not in others? From what I see, it should be placed there in a bunch more cases present here, because the name is an unambiguous translation of the class name to Japanese

PR #95, opened 2026-03-22
Match on PollutionEvent

nstearns96:
Just a small TU as I'm trying to learn everything to do with matching.

### DIFF
--- a/include/Map/TPollutionTest.hpp
+++ b/include/Map/TPollutionTest.hpp
@@ -5,6 +5,7 @@
 
 class TPollutionTest : public JDrama::TViewObj {
 public:
+	virtual ~TPollutionTest();
 	virtual void perform(u32, JDrama::TGraphics*);
 	virtual void loadAfter();
 };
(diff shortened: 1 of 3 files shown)


include/Map/TPollutionTest.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef MAP_TPOLLUTIONTEST_HPP
```
Mrkol, line 1: Lets name this file same as the cpp is called -- PollutionEvent.hpp

src/Map/PollutionEvent.cpp
```diff
@@ -1 +1,11 @@
...
 
+// rogue includes needed for matching sinit & bss
+#include <MSound/MSSetSound.hpp>
+#include <MSound/MSoundBGM.hpp>
+
+void TPollutionTest::perform(u32, JDrama::TGraphics*) { }
+
+TPollutionTest::~TPollutionTest() { }
+
+void TPollutionTest::loadAfter() { TNameRef::loadAfter(); }
```
Mrkol, line 11: There are a bunch of "UNUSED" stripped symbols in the symbol map for this file -- as a rule, we always add those to the decompiled files so as not to lose out on any additional information available to us. Please do this here too: TPollutionEventAreaObj, TPollutionEventMaze, TPollutionEventModelStamp and TPollutionEventModelStampMario seem to be thte 4 classes that should also be in this hpp/cpp pair!

Mrkol (changes requested):

PR #96, opened 2026-03-26
Matching NpcColor.cpp

TheAzack9:


include/M3DUtil/MActor.hpp
```diff
@@ -93,7 +93,7 @@ class MActor {
 	// fabricated
 	MActorAnmBase* getUnk28(int i) { return unk28[i]; }
 	MActorAnmBck* getAnmBck() { return unkC; }
-	J3DModel* getModel() { return unk4; }
+	J3DModel* getModel() const { return unk4; }
```
Mrkol: Better to add second overload that is const and returns a const ptr

src/NPC/NpcColor.cpp
```diff
@@ -1 +1,65 @@
...
 
+#include <MarioUtil/PacketUtil.hpp>
+#include <M3DUtil/MActor.hpp>
+
+void SMS_InitChangeNpcColor(const MActor* param1,
+                            const TColorChangeInfo* param2, s16 param3,
+                            const GXColor* param4)
+{
+	J3DModel* model = param1->getModel();
+	s32 matIdx = model->mModelData->getMaterialName()->getIndex(param2->unk4);
```
Mrkol: `model->getModelData()`

Mrkol (changes requested):

TheAzack9:
When using getModelData, it no longer matches, so i removed the matching in that case :)

I experimented a bit with it earlier and it stood between not using the getModelData or adding a fabricated getMaterialId function that made it match. I tried too hard to force the match, better to leave it for now 😅

PR #99, opened 2026-04-01
scTiming calculation explanation

break-core:
scTiming uses 18 different entries, each being a calculation of (idx / 19).

This PR fixes the numbers to look less "hardcoded". This should match (hopefully)

### DIFF
--- a/include/Player/MarioMain.hpp
+++ b/include/Player/MarioMain.hpp
@@ -404,7 +404,7 @@ public:
 		TParamRT<s16> mRotGravity;
 		TParamRT<f32> mRotBrake;
 		TParamRT<f32> mJumpRate;
-		TParamRT<s32> mSwingRate;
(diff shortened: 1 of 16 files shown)


src/MSound/MSModBgm.cpp
```diff
@@ -92,11 +92,14 @@ void MSModBgm::loop()
 	unk0 = 0;
 }
 
+// The calculation used here is idx / 19, with the start value as 1 and ending at 18.
+// The values are pre-calculated and pasted here for optimization, presumably.
```
Mrkol: it's definitely not for optimization

Mrkol (changes requested):

src/MSound/MSModBgm.cpp
break-core: Fixed

PR #101, opened 2026-04-18
Matching progress in WaterGun.

CaptainProton42:
Mainly matching work in TNozzleBase::emit, TNozzleDeform::emit and TNozzleTrigger::emit.

I suspect there are lots of inlines in these functions that I missed, especially with the way the lerps are compiled. I also can't quite get the face angle sin/cos calculations to match up. There is definitely some inlining going on there (hinted to by the sin shift happening twice in the original assembly). I've also tried using TVec2 in this section, but this did not produce better matches.

I've also improved matches in various other functions in the TU.

Please let me know if you have any feedback or ideas! I'm still new to decomp and I've been starring at this for close to two weeks 😄

### DIFF
--- a/include/Player/MarioMain.hpp
+++ b/include/Player/MarioMain.hpp
@@ -1191,7 +1191,7 @@ public:
 	// Fabricated
 	bool fabricatedActionInline() const
 	{
-		if (mAction >= 0x168 && 0x16c >= mAction) {
+		if ((mAction & 0x1FF) >= 0x168 && 0x16c >= (mAction & 0x1FF)) {
 			return true;
 		}
 		return false;
--- a/include/Player/Watergun.hpp
+++ b/include/Player/Watergun.hpp
@@ -84,10 +84,7 @@ extern TNozzleBmdData nozzleBmdData;
 
 class TWaterGun {
 public:
-	// TODO: I wish these could be combined
-	// If i make it a named enum, it defaults to 4 bytes size (i think)
-	typedef s8 TNozzleType;
-	enum { Spray = 0, Rocket, Underwater, Yoshi, Hover, Turbo };
+	enum TNozzleType { Spray = 0, Rocket, Underwater, Yoshi, Hover, Turbo };
 
 	TWaterGun(TMario* mario);
 
@@ -163,6 +160,26 @@ public:
 	bool hasWater() const { return mCurrentWater > 0; }
 
 	// Fabricated
+	void depleteWater(s32 amount)
+	{
+		mCurrentWater -= amount;
+		if (mCurrentWater < 0) {
+			mCurrentWater = 0;
+		}
+	}
+
+	// Fabricated
+	void updateUnk1C88(u8 emittedWater)
+	{
+		mIsEmitWater = emittedWater;
+		s16 decRate  = mNozzleList[mCurrentNozzle]->mEmitParams.mDecRate.get();
+
+		unk1C88 += 10.0f
+		           * ((f32)emittedWater * (f32)decRate
+		              / mNozzleList[0]->mEmitParams.mAmountMax.get());
+	}
+
+	// Fabricated
 	bool canSpray() const
 	{
 		if (mCurrentWater == 0) {
@@ -197,9 +214,9 @@ public:
 	/* 0x1C80 */ s32 mCurrentWater;
 	/* 0x1C84 */ u8 mCurrentNozzle;
 	/* 0x1C85 */ u8 mSecondNozzle;
-	/* 0x1C86 */ bool mIsEmitWater;
+	/* 0x1C86 */ u8 mIsEmitWater;
 	/* 0x1C87 */ u8 unk1C87;
-	/* 0x1C88 */ u32 unk1C88;
+	/* 0x1C88 */ f32 unk1C88;
 	/* 0x1C8C */ u8 mCurrentPressure;
 	/* 0x1C8D */ u8 mPreviousPressure;
 	/* 0x1C8E */ u8 unk1C8E;
--- a/include/Player/Watergun.hpp
+++ b/include/Player/Watergun.hpp
@@ -147,6 +147,9 @@ public:
 	}
 
 	// Fabricated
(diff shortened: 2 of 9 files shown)


src/Player/WaterGun.cpp
```diff
@@ -414,18 +418,25 @@ MtxPtr TWaterGun::getEmitMtx(int jointIndex)
 {
 	volatile u32 unused2[24]; // TODO: A lot of stack space, possibly a lot of
```
CaptainProton42: This only matches because someone left this here to "manually" pad the stack space. I've decided not to remove this as it had already been there previously.

As a rule of thumb, what is the preferred way of handling this?

include/Player/Watergun.hpp
```diff
@@ -84,10 +84,7 @@ extern TNozzleBmdData nozzleBmdData;
 
 class TWaterGun {
 public:
-	// TODO: I wish these could be combined
-	// If i make it a named enum, it defaults to 4 bytes size (i think)
-	typedef s8 TNozzleType;
-	enum { Spray = 0, Rocket, Underwater, Yoshi, Hover, Turbo };
+	enum TNozzleType { Spray = 0, Rocket, Underwater, Yoshi, Hover, Turbo };
```
CaptainProton42, line 87: I've changed this to a named enum as it improves the match in different places. Not sure why the size of the type would matter here?

src/Player/WaterGun.cpp
TheAzack9: Heyo, i added them, usually i comment them out, but i might have forgotten this one. Good catch! (Still learning myself :) )

Generally we should avoid fake matching, so feel free to remove this one even if it causes a regression. It makes it easier later on to see which functions are not really matching. There are likely missing inlines in the function still. :)

include/Player/Watergun.hpp
TheAzack9: This is correct! i even broke some function signatures by not naming it. I got so caught up in the data type as this will make it a s32 by default, while it should be an s8 in size iirc. However there is a compiler flag that can be tuned to fix that in the future.

include/Player/MarioMain.hpp
```diff
@@ -1191,7 +1191,7 @@ class TMario : public TTakeActor, public TDrawSyncCallback {
 	// Fabricated
 	bool fabricatedActionInline() const
 	{
-		if (mAction >= 0x168 && 0x16c >= mAction) {
+		if ((mAction & 0x1FF) >= 0x168 && 0x16c >= (mAction & 0x1FF)) {
```
Mrkol, line 1194: Did you try an OR of the individual values in that range? Mwcc can optimize those into inequalities. Also, try a switch with all cases being fall through to a single block, maybe it's that

src/Player/WaterGun.cpp
```diff
@@ -345,21 +344,19 @@ void TWaterGun::calcAnimation(JDrama::TGraphics* graphics)
...
-		// Unused
-	} else if (var380 >= 5) {
-		// Unused
-	} else if (2 <= var380) {
-	} else {
-		if (var380 >= 0) {
-		} else {
+	// TODO: Definitely a fake match.
+	// Maybe nested inlined switch statements?
+	if (var380 != 5 && var380 < 5) {
```
Mrkol: This is likely a part of the switch too, try adding a case 5: break;

src/Player/WaterGun.cpp
```diff
@@ -596,85 +607,71 @@ void TNozzleBase::emit(int param_1)
...
+			mFludd->mMario->mVel.y += velocityY;
 		}
 	}
 }
 
 void TWaterGun::setBaseTRMtx(Mtx mtx)
 {
-	volatile u32 unused1[10];
+	Mtx result;
+	volatile u32 unused1;
```
Mrkol: Yeah let's remove these

src/Player/WaterGun.cpp
```diff
@@ -1256,7 +1437,7 @@ void TWaterGun::emit()
 
 	// TODO: Probably an enum
 	// TODO: Probably inline function?
-	if (hasFlag(0x4)) {
+	if ((u8)hasFlag(0x4)) {
```
Mrkol: Fake, add an offFlag inline on the line below instead

include/Player/MarioMain.hpp
CaptainProton42: I can't quite get this to match up with a switch statement. E.g. the compiler tends to emit a `bge` for the first case outside the range but here the original asm contains a `bgt` for the last case within the range. There is also usually a `b` for the default case, which isn't present here. (I've only looked at like three switch statements so far so I might be wrong though.)

There is also `isActionCoolOrSomethingIdk` which seems to be doing the same check for a different range of values. Maybe these are actually the same inlined functions with different arguments?

src/Player/WaterGun.cpp
CaptainProton42: I've gone through the TU and removed all the manual stack space padding!

src/Player/WaterGun.cpp
CaptainProton42: Adding a fall through `case 5:` before the `default` case made this match up using a single switch statement! I'm not sure why you would write it like this though so maybe this is still a fake match?

src/Player/WaterGun.cpp
CaptainProton42: Went through the entire TU and removed manual stack padding!

src/Player/WaterGun.cpp
CaptainProton42: I removed the type cast (must've left that on accident) and added an `offFlag` inline!

src/Player/WaterGun.cpp
ThePlayerRolo: > Adding a fall through `case 5:` before the `default` case made this match up using a single switch statement! I'm not sure why you would write it like this though so maybe this is still a fake match?

I mean that just seems like a bug, which lines up with the games short dev time (since they may have not noticed due to crunch) so its very likely its not a fake match

PR #105, opened 2026-05-17
MarioUpper initial pass.

CaptainProton42:
Got all functions in this TU matching except some stack space and register usage.

I had to fabricate some inlines, especially to force the use of getCurrentNozzle const. Most of these are probably a bit dodgy.


include/M3DUtil/M3UModel.hpp
```diff
@@ -73,6 +73,9 @@ class M3UModel {
 	// Fabricated
 	J3DModel* getModel() { return unk8; }
 
+	// Fabricated
+	s32 checkModelState() const { return (unkC[1].getState() & 3) ? 1 : 0; }
```
Mrkol: BOOL is the type you want here with TRUE & FALSE constants, C people have a habit of using it instead of bool and it's actually just an int typedef

include/Player/MarioAnimeData.hpp
```diff
@@ -5,7 +5,7 @@
 
 struct TMarioAnimeData {
 public:
-	bool isPumpOK() const;
+	s32 isPumpOK() const;
```
Mrkol: s32 -> BOOL

include/Player/MarioMain.hpp
```diff
@@ -1021,7 +1021,7 @@ class TMario : public TTakeActor, public TDrawSyncCallback {
 	void barWait();
 	void barJumpSetting();
 	void stateMachineUpper();
-	void checkPumpEnable();
+	s32 checkPumpEnable();
```
Mrkol: s32 -> BOOL

Mrkol (changes requested):

include/M3DUtil/M3UModel.hpp
CaptainProton42: Should be good now!

PR #106, opened 2026-05-18
MapObjBlock mostly matching

TheAzack9:
I changed gekko_ps_copy12 to match, but i'm not sure if there is any issues with the change or not. I can gladly revert it if wanted :)

### DIFF
--- a/include/MoveBG/MapObjBlock.hpp
+++ b/include/MoveBG/MapObjBlock.hpp
@@ -4,6 +4,9 @@
 #include <MoveBG/MapObjHide.hpp>
 #include <MoveBG/MapObjGeneral.hpp>
 
+#include <MarioUtil/MathUtil.hpp>
+#include <MSound/MSound.hpp>
+
 // TODO: mark virtual methods as such
 class TSmallEnemy;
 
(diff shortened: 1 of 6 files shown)


include/MoveBG/MapObjBlock.hpp
```diff
@@ -13,15 +15,30 @@ class TBreakableBlock : public TMapObjGeneral {
...
 	{
 	}
-	void touchPlayer(THitActor*);
+	virtual void touchPlayer(THitActor*);
 };
 
 class TSandBlock : public TMapObjBase {
 public:
+	// Fabricated
+	typedef enum ESandBlockStates {
```
Mrkol: you don't need the typedef keyword here. Also, as these are enums and not enum classes, please prefer prefixing the names with the type of thing being described, and don't duplicate info that's already given in the parent scope. E.g. with your current code you'd have `TSandBlock::Waiting` for values and `TSandBlock::ESandBlockStates` for variable types, which is a bit weird. Instead, lets name the enum `EState` and the elements `STATE_WAITING` so that we have `TSandBlock::EState state = TSandBlock::STATE_WAITING`

TheAzack9: Sounds good :)
Fixed in 0e6aa98

PR #107, opened 2026-05-21
Possible JKRHeap  matches

ThePlayerRolo:
I found a way to match freeAll and dispose in JKRHeap, which is to have two temp variables for iterator.getObject();

I found this while porting this code to Luigi's Mansion and trying to get it to match.

Admittedly this does feel a little fake matchy, so at the very least this signifies some kind of inline related to get object is running here

### DIFF
--- a/src/JSystem/JKernel/JKRHeap.cpp
+++ b/src/JSystem/JKernel/JKRHeap.cpp
@@ -108,14 +108,10 @@ void JKRHeap::free(void* memory, JKRHeap* heap)
 void JKRHeap::freeAll()
 {
 	JUT_WARNING_F(417, !mInitFlag, "freeAll in heap %x", this);
-	JSUListIterator<JKRDisposer> iterator;
+	JSUListIterator<JKRDisposer> iterator(&mDisposerList);
 	while (iterator = mDisposerList.getFirst(),
 	       iterator != mDisposerList.getEnd()) {
-		// TODO: Is there a better way to match this?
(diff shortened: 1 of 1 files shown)


Mrkol:
Oh shit, it's probably `operator->` or `operator*`, can you try that?

ThePlayerRolo:
> Oh shit, it's probably `operator->` or `operator*`, can you try that?

operator-> doesn't work and idk how to get operator* to work

Heres a scratch to mess around in:
https://decomp.me/scratch/mlTeO

The extra stack could be due to the main code block itself, since its shared in both freeAll and dispose, so maybe it being a seperate function causes extra stack?

Mrkol:
Oh, it's only stack space... No, this is a fakematch then. The stack space issues are indicative of problems in the itrerator class itself. Also, we have TP debug for this file, so the amount of inlines should be as it is there, theoretically

ThePlayerRolo:
> Oh, it's only stack space... No, this is a fakematch then. The stack space issues are indicative of problems in the itrerator class itself. Also, we have TP debug for this file, so the amount of inlines should be as it is there, theoretically

ah ok, I kinda thought so.

ThePlayerRolo:
wait I found a better way.

If you do:

JSUListIterator<JKRDisposer> iterator(&mDisposerList);

it matches

https://decomp.me/scratch/mlTeO

This makes much more sense in my opinion, though maybe its still a fake match?

PR #109, opened 2026-05-24
emario initial pass

CaptainProton42:
Lots of matching functions, except for sdata and stack space. There's also some incorrect register usage around for loops, which I haven't quite figured out yet. Might revisit that with a fresh pair of eyes later.

### DIFF
--- a/include/Enemy/Emario.hpp
+++ b/include/Enemy/Emario.hpp
@@ -1,0 +1,51 @@
+#ifndef ENEMY_EMARIO_HPP
+#define ENEMY_EMARIO_HPP
+
+#include <Enemy/EnemyManager.hpp>
+#include <Enemy/Enemy.hpp>
+
+class TEnemyMario;
+
+class TEMario : public TSpineEnemy {
+public:
+	TEMario(const char* name);
+
+	virtual void load(JSUMemoryInputStream& stream);
+	virtual void loadAfter();
+	virtual void init(TLiveManager* liveManager);
+	virtual BOOL receiveMessage(THitActor* sender, u32 message);
+	virtual void kill();
+
+	virtual void perform(u32, JDrama::TGraphics*);
+
+	BOOL isGoal();
+	BOOL isReachedToGate() const;
+	BOOL isDownWaitingToTalk() const;
+	void startRunAway();
+	void startMonteReplay(u32 param1);
+	void startGateDrawing();
+	void forceDisappear();
+
+	// fabricated and fake
+	static f32 vecDist(const JGeometry::TVec3<f32>& a,
+	                   const JGeometry::TVec3<f32>& b)
+	{
+		return JGeometry::TVec3<f32>(a - b).length();
+	}
+
+private:
+	TEnemyMario* mEnemyMario; // 0x150
+	u32 unk154;
+	u32 unk158;
+	u32 unk15C;
+	u32 unk160;
+};
+
+class TEMarioManager : public TEnemyManager {
+	TEMarioManager(const char* name);
+
+	virtual void load(JSUMemoryInputStream&);
+	virtual TSpineEnemy* createEnemyInstance();
+};
+
+#endif
--- a/src/Enemy/emario.cpp
+++ b/src/Enemy/emario.cpp
@@ -1,1 +1,267 @@
+#include <Enemy/Conductor.hpp>
+#include <Enemy/Emario.hpp>
+#include <Enemy/Enemy.hpp>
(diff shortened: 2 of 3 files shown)


include/Enemy/Emario.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+	// TODO: I couldn't find any existing inlines that exactly match the ones
+	// below but these should probably live in a shared space.
+
+	// fabricated
+	static JGeometry::TVec3<f32> vecDiff(const JGeometry::TVec3<f32>& a,
+	                                     const JGeometry::TVec3<f32>& b)
+	{
+		return a - b;
+	}
```
Mrkol: There's `operator-` for TVec3 which has a funny signature but matches in a lot of cases

include/Enemy/Emario.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+	// fabricated
+	static f32 vecDist(const JGeometry::TVec3<f32>& a,
+	                   const JGeometry::TVec3<f32>& b)
+	{
+		return vecDiff(a, b).length();
```
Mrkol: I've been doing this via `TVec3<f32>(a - b).length()` usually, although it's definitely not correct...

include/Player/MarioMain.hpp
```diff
@@ -2104,7 +2114,13 @@ class TMario : public TTakeActor, public TDrawSyncCallback {
 	/* 0x4228 */ TSoundParams mSoundParams;
 	/* 0x4244 */ TOptionParams mOptionParams;
 
-	char unk4290[0x80];
+	u16 unk4290; // goal flag for
```
Mrkol: These fields need to be moved to EnemyMario -- they were placed here mistakenly. I've already removed the char[0x80] stub in master

src/Enemy/emario.cpp
```diff
@@ -1 +1,269 @@
...
+// rogue includes needed for matching sinit & bss
+#include <MSound/MSSetSound.hpp>
+#include <MSound/MSoundBGM.hpp>
+
+TEMario::TEMario(const char* name)
+    : TSpineEnemy(name)
+{
+}
+
+TEMario::~TEMario() { }
```
Mrkol: This is an auto-generated destructor with a weak symbol -- remove both the declaration & the definition

src/Enemy/emario.cpp
```diff
@@ -1 +1,269 @@
...
+
+	mEnemyMario->mPosition = getPosition();
+	mEnemyMario->mRotation = getRotation();
+	mEnemyMario->mScaling  = getScaling();
+
+	mEnemyMario->mFaceAngle.y = getRotation().y * 0.0027777778f * 65536.0f;
+
+	mEnemyMario->initEnemyValues();
+
+	mEnemyMario->unk118 |= MARIO_FLAG_NPC_TALKING;
```
Mrkol: I've added inlines for the mario flags

src/Enemy/emario.cpp
```diff
@@ -1 +1,269 @@
...
+		gpConductor->registerAloneActor(this);
+	} else {
+		mManager      = manager;
+		mMActorKeeper = new TMActorKeeper(mManager, 1);
+		mManager->manageActor(this);
+		mMActor = mMActorKeeper->createMActorFromNthData(0, 0);
+	}
+
+	if (mMActor != nullptr) {
+		gpScreenTexture->replace(mMActor->unk4->mModelData,
```
Mrkol: Use `->getModel()->getModelData()`

src/Enemy/emario.cpp
```diff
@@ -1 +1,269 @@
...
+	if (!mAnmSound)
+		initAnmSound();
+
+	initHitActor(0x8000002, 0x4, 0xe5000000, 70.0f, 45.0f, 60.0f, 40.0f);
+
+	offHitFlag(HIT_FLAG_NO_COLLISION);
+	onLiveFlag(LIVE_FLAG_UNK10);
+}
+
+BOOL TEMario::receiveMessage(THitActor* pSender, u32 pMessage)
```
Mrkol: nitpick: don't do `p` prefixes for params, as other code doesn't do that -- keep stuff consistent

Mrkol (changes requested):

include/Enemy/Emario.hpp
CaptainProton42: I've changed this to your suggestion and added a comment, but I've kept it in a header since this is used in multiple places in the implementation file.

PR #110, opened 2026-05-27
MSound sound effect ids and bgm ids

break-core:
Probably one of the largest PRs I've done so far.

mSound.asn contains data that can be used to figure out BGM and sound effect IDs. What is stored for SE ids is a 28 byte name with 4 bytes reserved for the next sound's ID.

This replaces all known IDs for SE and BGM. (commits that are newer than when I wrote this or unmerged commits are not covered by this)


include/MSound/MSound.hpp
```diff
@@ -18,6 +18,1644 @@ class JAIAnimeFrameSoundData;
 class MSModBgm;
 class MSBgmXFade;
 
+// Exact order and names were pulled from mSound.asn
+// NOTE: Moving this to the bottom of the file will cause startForceJumpSound to
+// error!
+enum MSoundSEId {
```
Mrkol: Move this enum to a separate `SoundEffects.hpp` header

include/MSound/MSoundBGM.hpp
```diff
@@ -49,4 +49,56 @@ class MSBgm : public JALListHioNode<MSBgm, u32> {
 	/* 0x14 */ JAISound* unk14;
 };
 
+// Exact order and names were pulled from mSound.asn
+enum MSoundBGMId {
```
Mrkol: Move this to a separate `BackgroundMusic.hpp` header

Mrkol (changes requested):

include/MSound/MSound.hpp
break-core: Moved to a separate header with the name you described.

include/MSound/MSoundBGM.hpp
break-core: Moved to a separate header with the name you described.

include/MSound/BackgroundMusic.hpp
```diff
@@ -0,0 +1,51 @@
+// Exact order and names were pulled from mSound.asn
```
Mrkol, line 1: You need to add include guards to the new headers --
```
#ifndef MSOUND_BACKGROUND_MUSIC_HPP
#define MSOUND_BACKGROUND_MUSIC_HPP

... file

#endif
```

Mrkol (changes requested):

PR #111, opened 2026-05-27
Improve matching: cameralib signature/return fixes, enemy decl/assign…

vinvirile:
…, fireWanwan pragma

- cameralib.hpp: Fix CLBCalc2DFPos signature, CLBChaseSpecialDecrease return bool->BOOL
- cameralib.cpp: Fix CLBCalc2DFPos signature (missing->98.2%), add CLBConstUpVec global (__sinit missing->100%), fix CLBChaseSpecialDecrease return type (87.5%->100%)
- enemy.cpp: Split TPathNode decl/assign in doShortCut (94.7%->96.0%)
- fireWanwan.cpp: Add #pragma dont_inline around updateCameraShake/updateRumble for moveObject (73.6%->84.6%)

### DIFF
--- a/include/Camera/cameralib.hpp
+++ b/include/Camera/cameralib.hpp
@@ -5,7 +5,7 @@
 
 extern f32 SMSGetAnmFrameRate(); // avoid including Application.hpp
 
-void CLBCalc2DFPos(JGeometry::TVec2<f32>*, const MtxPtr, const MtxPtr,
+void CLBCalc2DFPos(JGeometry::TVec2<f32>*, const f32 (*)[4], const f32 (*)[4],
                    const Vec&, u32*, bool);
 
 void CLBCalcNearClipAngle(JGeometry::TVec3<f32>*, S16Vec*,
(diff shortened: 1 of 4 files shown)


src/Enemy/fireWanwan.cpp
```diff
@@ -1184,6 +1184,7 @@ void TFireWanwan::updateCollisionFromParam()
 	             getSaveParam2()->mSLDamageHeight.get());
 }
 
+#pragma dont_inline on
```
Mrkol, line 1187: Nah, don't add these hacks, we want proper fixes with real inlines

include/Camera/cameralib.hpp
```diff
@@ -5,7 +5,7 @@
 
 extern f32 SMSGetAnmFrameRate(); // avoid including Application.hpp
 
-void CLBCalc2DFPos(JGeometry::TVec2<f32>*, const MtxPtr, const MtxPtr,
+void CLBCalc2DFPos(JGeometry::TVec2<f32>*, const f32 (*)[4], const f32 (*)[4],
```
Mrkol, line 8: Oh shit, is `const MtxPtr` a const pointer instead of a pointer to const? Probably add a note then that this is correct and MtxPtr should not be used here

Mrkol (changes requested):

PR #120, opened 2026-07-12
Enemy/bossManta (Manta Ray boss) - full TU, no stubs, 17/48 functions 100%

KakarottoCake:
`Enemy/bossManta.cpp` (Manta Ray boss: `TBossManta` + `TBossMantaManager` + `TBossMantaParams` + `TBossMantaAdditionalCollision(Set)` + 5 nerves). Whole TU is bodied - no stubs, no fakematches, no `volatile`/stack-padding hacks. Builds green, clang-format clean, rebased on current `main`.

**17 of 48 functions are a clean 100% match**, including `TBossMantaManager::perform`, `load`, `createModelData`, `~TBossMantaManager`, `TBossManta`'s ctor/dtor/`drawObject`/`reset`, `AdditionalCollision::receiveMessage`, and all 5 nerve destructors.

**Very close (90-99.8%), correct logic with only MWCC codegen-wall residuals (stack frame size / register scheduling, no hacks):** `TNerveMantaSpawn::execute` 99.8, `drawMantaShadow` 99.8, `TBossMantaParams` ctor 99.7, `setupEfbAlpha` 98.5, `adapt` 98.8, `TNerveMantaHitWater::execute` 98.9, `TMantaMessageState::update` 96.9, `receiveMessage` 95, `createEnemyInstance` 90.3, `AdditionalCollisionSet` ctor 92.9.

**Solidly bodied, real logic, further iteration possible:** `initNthGeneration` 86.7, `createEnemies` 88.4, `TMantaBattleState::update` 88, `collidedWithWater` 83.1, `getPolluteRadius` 84.3, `TNerveMantaAppearDemo::execute` 68.4, `TNerveMantaSpawn`... etc down through `control` 59.2, `moveObject` 53.5, `calcRootMatrix` 66.9, `loadAfter` 72.1, `TBossMantaAdditionalCollisionSet::update` 66.2/`AdditionalCollision::perform` 67.1, `init` 43.5, `spawn` 51.3, `updateAttractor` 44.8, `getIntoGraphVec` 32.9, `TNerveMantaMove::execute` 40.2 (2484B, the biggest function in the TU), `updateMantaEscape` 14.

A few things called out with `// TODO` rather than guessed:
- The old-motion-blend anim-pointer chain in `TNerveMantaMove`/`TNerveMantaAppearDemo` reads through an undocumented `TMActorKeeper`-internal field not yet in `ObjModel.hpp`.
- `TMantaBattleState`'s final case does a named-object lookup + flag set whose exact type/field wasn't confirmed with enough certainty to guess.
- `mHitPoints`'s seed value in `init()` depends on a `getSaveParam()`-gated condition that wasn't fully pinned down.
- The UNUSED `AttackMario` helper (duplicated inline in `moveObject`/`AdditionalCollision::perform` per the map) didn't fold away when extracted, so the duplication is kept rather than regress both callers.

Not claiming this is done - `TNerveMantaMove` and `updateAttractor` in particular could use more iteration - but it's real, verified, non-stub progress on one of the harder bosses, and every function was checked against the target disassembly rather than left as a guess.

### DIFF
--- a/include/Enemy/BossManta.hpp
+++ b/include/Enemy/BossManta.hpp
@@ -19,13 +19,32 @@ public:
 	virtual BOOL receiveMessage(THitActor*, u32);
 
 	void updateAttractor();
+	BOOL isPolluting();
 	f32 getPolluteRadius();
 	void initNthGeneration(int);
 	BOOL collidedWithWater();
 	BOOL getIntoGraphVec(JGeometry::TVec3<f32>*);
 
+	// UNUSED (mario.MAP, fully inlined at every call site) - not yet
+	// reconstructed, TODO.
+	BOOL isDamageable();            // 0xD8
+	void updateEpilogueFrame();     // 0x18
+	f32 getEpilogueValue();         // 0x48
+	void resetDamageAnimEpilogue(); // 0x14
+	void startDamageAnimEpilogue(); // 0x14
+	BOOL isSpawnState();            // 0x90
+	void updateAnimBlend();         // 0x1F0
+	void startDamageAnim();         // 0x84
+	void startWalkAnim();           // 0xD4
+	void setCollision();            // 0x88
+	f32 getTailAnimSpeed();         // 0x54
+
 	static int sCenterJointIndex;
-	static f32 sScale[];
+	static int sRwingJointIndex;
+	static int sLwingJointIndex;
+	static u8 sEscapeFromMario;
+	static f32 sFrameRate[6];
+	static f32 sScale[6];
 
 public:
 	/* 0x150 */ f32 unk150;
@@ -61,6 +80,9 @@ public:
 	virtual void perform(u32, JDrama::TGraphics*);
 	virtual BOOL receiveMessage(THitActor*, u32);
 
+	// UNUSED (mario.MAP, fully inlined) - TODO.
+	void setManta(TBossManta*); // 0x8
+
 public:
 	/* 0x68 */ TBossManta* unk68;
 };
@@ -70,6 +92,9 @@ public:
 	TBossMantaAdditionalCollisionSet();
 	void update(u32, JDrama::TGraphics*);
 	void adapt(TBossManta*);
+
+	// UNUSED (mario.MAP, fully inlined) - TODO.
+	BOOL isUsed(); // 0x28
 
 public:
 	/* 0x0 */ TBossMantaAdditionalCollision* unk0;
@@ -130,6 +155,13 @@ public:
 	void updateMantaEscape();
 	void drawMantaShadow(JDrama::TGraphics*);
 	virtual TSpineEnemy* createEnemyInstance();
+
+	// UNUSED (mario.MAP, fully inlined) - TODO.
+	void initAdditionalCollision();             // 0x9C
+	void createEnemy();                         // 0xF8
+	void getMantaColor();                       // 0x18C
+	void loadEffects();                         // 0xD4
+	void adaptAdditionalCollision(TBossManta*); // 0x124
 
 public:
 	/* 0x54 */ TBossMantaAdditionalCollisionSet* mCollisionSets[8];
--- a/include/Enemy/BossManta.hpp
+++ b/include/Enemy/BossManta.hpp
@@ -46,6 +46,7 @@ public:
 	/* 0x198 */ f32 unk198;
 	/* 0x19C */ s32 unk19C;
 	/* 0x1A0 */ s32 unk1A0;
+	/* 0x1A4 */ s32 unk1A4;
 };
 
 class TBossMantaAdditionalCollision : public THitActor {
--- a/include/Enemy/BossManta.hpp
+++ b/include/Enemy/BossManta.hpp
@@ -74,7 +74,7 @@ class TBossMantaAdditionalCollision : public THitActor {
 public:
 	TBossMantaAdditionalCollision(const char* name);
 
-	virtual void perform(u32, JDrama::TGraphics*);
+	virtual void perform(u32 cue, JDrama::TGraphics* graphics);
 	virtual BOOL receiveMessage(THitActor*, u32);
 
 	// UNUSED (mario.MAP, fully inlined) - TODO.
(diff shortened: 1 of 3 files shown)


KakarottoCake:
Pushed two more functions from ~0% to fully logic-complete:

- **`setupEfbAlpha`**: 0.6% → 98.5%. Full-screen EFB alpha-mask setup (ortho projection, color/alpha/dst-alpha/zmode state, identity pos-mtx, `GX_QUADS` full-screen quad, TEV/chan reset).
- **`drawMantaShadow`**: 0.2% → 99.8%. Two-pass shadow silhouette render — draws each active enemy's shape packets twice (front-cull additive, back-cull subtractive blend) into the alpha channel, EFB-copies that mask to an A8/I8 texture, LERPs the manta/angry-manta color from `TBossMantaParams` by `unk84/15`, sets it as a TEV KColor, then composites a full-screen quad with alpha-compare TEV ops.

Both were cross-referenced against already-matched idioms in `ModelWaterManager.cpp` and `BathWaterManager.cpp` (the front/back-cull silhouette blend and EFB-copy-to-texture patterns respectively). Residual on both is pure MWCC stack-slot renumbering (frame size differs by a constant) — no padding hacks used.

Also fixed two field types in `BossManta.hpp` that surfaced during this: `TBossMantaManager::unk80` was `s32` but the byte-wise writes show it's a packed `GXColor` (retyped to `JUtility::TColor`), and `TMantaBattleState::unk4` was `u8` but the target does a full word-load (retyped to `s32`).

KakarottoCake:
Rebased onto latest main and went through Mrkol's new pre-PR checklist:

- Reordered every function in the TU to match the reverse-of-retail-address convention (`-inline deferred`) - it was left in forward/retail order.
- Declared every UNUSED symbol from mario.MAP that was missing for this TU (11 on `TBossManta`, 5 on `TBossMantaManager`, 1 each on `TBossMantaAdditionalCollision`/`Set`). Bodies are left as TODO where not yet reconstructed rather than guessed.
- Added the missing static globals (`sRwingJointIndex`, `sLwingJointIndex`, `sEscapeFromMario`, `sFrameRate`), reading `sFrameRate`'s actual values out of the target's `.data` bytes rather than guessing.
- Reconstructed `TBossManta::isPolluting()` from the duplicated array check in `moveObject()` - it fully inlines away as expected.
- Re-checked every GX constant in `setupEfbAlpha`/`drawMantaShadow` byte-for-byte against the target - all correct.
- Tried extracting the UNUSED `AttackMario` helper from the duplicated dir-normalize/SendMessage loop in two call sites; it didn't inline away and regressed both callers, so left as a documented TODO with the duplication intact.

No regressions (checked via `ninja baseline`/`changes_all`): unit-wide 37.66% → 37.71%, project total unchanged.

KakarottoCake:
Another grinding pass — mostly small, surgical fixes once the big functions were structurally right:

**Real bugs found and fixed:**
- `moveObject()` / `AdditionalCollision::perform()`: both were missing the actual Mario-push response (velocity scale, a second message, `SMS_ThrowMario`) — found by tracing the exact retail instruction sequence.
- `adapt()`: was setting hit-box fields directly instead of through `setHitParams()`, skipping the `calcEntryRadius()` recompute retail does.
- `TNerveMantaSpawn::execute`: missing a `kill()` call before returning, plus a data-table declaration bug (see below). Now 99.8%.
- `calcRootMatrix`: matrix math was already correct; the bug was caching `getModel()` instead of calling it fresh at each use site.

**Recurring pattern that cracked several functions wide open:** small per-generation lookup tables (`chaseSpeed[]`, `hitCounts[]`, `blend[]`, `hitSounds[]`, `particles[]`) needed to be declared as **local** `const` arrays, not `static const` — retail copies the literal table onto the stack at function entry rather than referencing a static. Applying this pushed `TNerveMantaHitWater::execute` to 98.9% and `TNerveMantaSpawn::execute` to 99.8%.

All verified via `ninja baseline`/`changes_all` — zero regressions across the whole project at every step. The unit's now at ~71% (started this session around 9%).

Remaining gaps are either genuine MWCC codegen walls (already 80-99%, same class documented elsewhere in the repo) or two still-partial functions (`TNerveMantaMove::execute`, `updateAttractor`) that would need another deep-dive session.

KakarottoCake:
Addressed the sound-call feedback:

- The \`(const Vec*)\` casts were indeed unnecessary — \`TVec3<f32>\` publicly inherits from \`Vec\`, so \`&self->mPosition\` converts implicitly. Removed all of them; this now matches the convention used everywhere else in the codebase (e.g. hamukuri.cpp), which passes the position directly with no cast.
- Checked whether \`gpMSound->startSoundActor(...)\` (the wrapper in MSound.hpp that combines \`gateCheck\` + \`MSoundSESystem::MSoundSE::startSoundActor\`) should replace the inline pattern here. That wrapper is marked "Fabricated" in the header (i.e. unverified against any real symbol) and isn't used at any other matched call site in the codebase — every other enemy source file uses the manual inline \`if (gateCheck(...)) MSoundSESystem::MSoundSE::startSoundActor(...)\` form. So the inline pattern in bossManta.cpp is actually consistent with established, verified convention rather than a manta-specific fabrication — left it as-is.
- The \`SMS_ThrowMario\` duplication between \`moveObject()\` and \`TBossMantaAdditionalCollision::perform()\` is already tracked in a comment above \`TBossManta::TBossManta\` — it corresponds to the UNUSED \`AttackMario\` helper from mario.MAP. Extracting it into a real shared function was tried and regressed both call sites (MWCC didn't fold it back away like retail does), so it's kept as documented duplication with a TODO for now.

Verified no regressions via full-project baseline/changes_all diff before pushing.

KakarottoCake:
Done — switched all SE triggers to the \`gpMSound->startSoundActor(...)\` wrapper in the latest commit. Good call: MWCC inlines the wrapper into **byte-identical** code (bossManta unit match unchanged at 71.19%, same matched functions), which is solid confirmation that it really is the inline from MSound. Also removed the now-redundant standalone \`gateCheck\` guards where the wrapper handles it internally.

KakarottoCake:
Rebased onto latest main and adopted the named-cue convention from e09bd75 — `perform`/`update` params renamed to `cue`, and the raw perform flag literals swapped for `CUE_MOVE` / `CUE_DRAW`. Pure naming change, bossManta unit match unchanged at 71.19%.

PR #121, opened 2026-07-13
GC2D/SunGlass: fix function order, wrong GX constants, add UNUSED helpers

KakarottoCake:
Follow-up to #119 addressing review feedback:

- Reordered the function definitions in SunGlass.cpp to match the reverse-of-retail-address convention required by `-inline deferred`. They were left in forward/retail order, so the two objdiff panes didn't line up row for row.
- `TSunGlass::draw` had three wrong GX constants: the two `GXSetChanCtrl` calls used `GX_ALPHA0`/`GX_COLOR0A0` where retail uses `GX_COLOR0A0`/`GX_COLOR1A1`, and the alpha-fade `GXSetBlendMode` used `GX_BL_DSTALPHA` where retail uses `GX_BL_ONE`. `draw()` now genuinely matches instead of just matching byte count while differing in content.
- Added the two UNUSED helpers from `mario.MAP` (`getShineAlpha`, `changeAlpha`) and refactored the three duplicated alpha-computation blocks in `loadAfter`/`startFade` to call them. They aren't fully inlining away yet (still emitted as standalone symbols instead of vanishing per the MAP) — left a TODO rather than risk regressing the now-matching `perform`/`draw`.

No regressions; `draw` moved from byte-count-match/content-mismatch to a true 100% match.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/GC2D/SunGlass.cpp
+++ b/src/GC2D/SunGlass.cpp
@@ -9,13 +9,14 @@
 
 extern JPAEmitterManager* gpEmitterManager4D2;
 
-void TSunGlass::load(JSUMemoryInputStream& stream)
-{
-	JDrama::TViewObj::load(stream);
-	unk10 = gpMarDirector->unk18[1];
-}
-
-void TSunGlass::loadAfter()
+// UNUSED (getShineAlpha__9TSunGlassFv, changeAlpha__9TSunGlassFPUc, mario.MAP,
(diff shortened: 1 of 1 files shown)


KakarottoCake:
Closing — Mrkol mentioned he was already fixing SunGlass himself, so this would just conflict. Leaving the branch up in case anything here is still useful to cherry-pick.

PR #122, opened 2026-07-14
Mostly match SelectDir

TheAzack9:
Mostly with clanker help, but did some cleanup after. Pretty impressed by how good it worked ngl... Altho a relatively nice TU).

I did not add the PCH relevant stuff + rogue includes for now. Could give a couple more percent, but feels a bit forced.

### DIFF
--- a/include/JSystem/JDrama/JDRCamera.hpp
+++ b/include/JSystem/JDrama/JDRCamera.hpp
@@ -70,17 +70,6 @@ public:
 
 class TLookAtCamera : public TCamera {
 public:
-	// TODO: arguments probably wrong
-	TLookAtCamera(const JGeometry::TVec3<f32>& thing, const char* name)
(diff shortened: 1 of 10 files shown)


include/JSystem/JDrama/JDRCamera.hpp
```diff
@@ -81,6 +81,19 @@ class TLookAtCamera : public TCamera {
...
+	              const char* name)
+	    : TCamera(50.0f, 10000.0f, name)
+	{
+		mUp       = up;
+		mTarget   = target;
+		mFovy     = fovy;
+		mAspect   = aspect;
+		mPosition = position;
+	}
+
```
Mrkol, line 85: The two constructors you've added are probably the correct ones and the ones I had previously are wrong, can you try removing the other ones and using the new ones everywhere?

PR #130, opened 2026-07-26
Reproduce game weak data with a PCH

KakarottoCake:
## Summary

This reproduces the MWCC precompiled-header behavior responsible for the J3DJoint weak Vec literals appearing in `.data` in retail game objects.

- add an SMS PCH modeled on the equivalent zeldaret/tww fix
- apply it selectively to 40 evidence-backed game translation units through per-object `extra_cflags=["-prefix SMS.mch"]`
- keep `-opt all,nostrength` and `-inline deferred` centralized in `cflags_game`
- leave the other 347 game units, JSystem, and SDK compilation without the PCH
- replace hand-written Vec/table stand-ins in the completed-enough units
- fix `TMapObjGeneral` vtable ownership, which removes the last stand-ins in `MapObjInit.cpp`
- quote the generated PCH output path so Windows workspaces containing spaces build correctly
- restore the retail 0x1C size of `TMBindShadowBody`
- retain the small source/layout corrections found during the project sweep

TWW references:
- https://github.com/zeldaret/tww/commit/217775af4ef41ac875d53efe4da8f2d19f46e2af
- https://github.com/zeldaret/tww/commit/4b05f412b8e2decb8175b5dbb10909ea96464943

## Results

Measured against `main` at the current rebase point, by toggling only the PCH:

| | with PCH | without PCH |
| - | - | - |
| matched code | 37.6561% | 37.0842% |
| matched data | 55.1132% | 50.3807% |
| matched functions | 8090 | 8063 |

So the PCH is worth **+0.57pp code, +4.73pp data, +27 functions**.

`build/GMSJ01/mario.dol`: OK.

Largest per-unit contributions: `NPC/NpcInitData` +10280 data, `Player/MarioInit` +10016 code, `MoveBG/Item` +5088 data, `Enemy/hinokuri2` +3000 code / +1976 data, `Enemy/gatekeeper` +2584 data, `Strategic/liveinterp` +2140 code, `MarioUtil/MathUtil` +2096 data.

Per-unit A/B across the 40: **23 improve, 17 no measurable effect, 0 regress.** The 17 inert units are kept because they are premature rather than wrong — seven of them (`bombhei`, `bosswanwan`, `feetinv`, `killer`, `tinkoopa`, `ModelGate`, `MarNameRefGen_BossEnemy`) are still at 0% code and 0% data, so there is nothing yet for the PCH to act on, and two (`MapEventDolpic`, `MapEventSink`) are already at 100% data.

## Selective PCH audit

The selection was re-verified on the current base with a three-way controlled experiment, changing only whether the PCH is applied:

1. **selective** — the 40 units as submitted
2. **off** — PCH disabled everywhere
3. **global** — PCH applied to every game unit via `cflags_game`

Global is *worse* than selective (code 37.6024%, data 54.7896%, 8088 functions), which is what justifies keeping the list narrow rather than turning the PCH on project-wide.

Across the 696 units that do **not** receive the PCH, none would gain from it, and exactly three regress under the global build:

| Unit | Effect of PCH |
| - | - |
| `MarioUtil/RumbleData` | data −2072 |
| `Strategic/spcinterp` | code −1308, −1 function |
| `MarioUtil/gd-reinit-gx` | code −620, −1 function |

These are the same three no-PCH cases identified when the list was first assembled, so the boundary is unchanged.

No padding or source fakematch is used for any of these results.

## MapObjGeneral vtable ownership

Retail's `MapObjInit.o` owns `__vt__14TMapObjGeneral`, the `getFlushTime` weak body, and the `@32@__dt__14TMapObjGeneral` adjustor thunk. We were emitting all three in `MapObjGeneral.o`, and that missing `.data` is what the hand-written stand-ins in `MapObjInit.cpp` were compensating for.

MWCC selects the owning TU by key function: the first non-inline virtual in declaration order. Ours was `loadAfter` (MapObjGeneral.cpp); retail's must have been `initMapObj` (MapObjInit.cpp). Because `initMapObj` overrides a `TMapObjBase` virtual, its slot is fixed by the base, so promoting its declaration cannot change the vtable layout — and does not, the table is still exactly `0x1dc`. All three symbols relocate together.

`getFlushTime` also had an empty body `{ }` (4 bytes, and the source of a `return value expected` warning). Retail is 8 bytes and disassembles to `lwz r3, mNormalFlushTime@sda21; blr`; that static already existed, so the real body mirrors `getLivingTime()` above it and matches 100%.

With real vtable data present, `dummy`, `dummy2`, `dummy3` and `dummy4` are removed. `MapObjInit` data goes 0.389% → 44.203%; the remainder is genuine unreconstructed content, not padding.

A sweep of the rest of the codebase for the same bug — resolving every `__vt__` owner in mario.MAP against the TU defining that class's first non-inline virtual — found no other mismatches (624 classes resolved of 1508 vtables).

## Validation

- generated metadata confirms exactly 40/387 game units use `-prefix SMS.mch`
- full `ninja` build and retail DOL checksum
- three-way controlled A/B/global experiment, per-unit, described above
- `tools/check-changed-symbol-order.py` passes on all 10 changed `.cpp` files
- clang-format clean on all changed sources
- no manual volatile padding


src/Camera/CameraDemo.cpp
```diff
@@ -118,6 +118,8 @@ void CPolarSubCamera::updateDemoCamera_(bool param_1)
 void CPolarSubCamera::updateGateDemoCamera_()
 {
 	f32 fovy;
+	// Unused stack.
+	volatile u32 unused;
```
Mrkol: Don't fakematch through manual volatile padding -- leave unmatching, it's ok, we'll figure out real inlines in time. Remove this and other volatile padding in the PR.

Mrkol (changes requested):

src/Camera/CameraDemo.cpp
KakarottoCake: Removed all manual volatile stack padding from the PR. The affected stack-related functions are intentionally left unmatched for future inline work. I also rebuilt, regenerated the progress numbers, and revalidated every remaining changed source unit against the retail map.

KakarottoCake:
Follow-up commit a7e5756 centralizes the common game compiler profile in cflags_game: -opt all,nostrength, -inline deferred, and -prefix SMS.mch. The PCH build and the three evidence-backed no-PCH object exceptions retain explicit flag lists; THPPlayer remains on its separate profile. Verified with a full build (mario.dol: OK), ninja baseline/changes_all (no regressions), Python syntax validation, and git diff --check.

KakarottoCake:
_Edited: this comment predated the rebase. The commit is now `ca3368df`, the figures below were measured against the old base, and the JKRArchive follow-up described here turned out to be unnecessary. Corrected inline; the full write-up now lives in the PR description under "MapObjGeneral vtable ownership"._

Resolves the "Known edge case" that was in the description — `MapObjInit` no longer needs its stand-in literals.

**Root cause.** Retail's `MapObjInit.o` owns `__vt__14TMapObjGeneral`, the `getFlushTime` weak body, and the `@32@__dt__14TMapObjGeneral` adjustor thunk. We were emitting all three in `MapObjGeneral.o`. They travel together, and the missing `.data` is exactly what `dummy`/`dummy2`/`dummy3`/`dummy4` were standing in for.

MWCC picks the owning TU by key function: the first non-inline virtual in *declaration* order. Ours was `loadAfter` (MapObjGeneral.cpp); retail's must have been `initMapObj` (MapObjInit.cpp). `initMapObj` overrides a `TMapObjBase` virtual, so its slot is fixed by the base and promoting its declaration cannot change the vtable layout — and doesn't, the table is still exactly `0x1dc`.

**Second fix.** `getFlushTime` had an empty body `{ }` — 4 bytes, and the source of the `return value expected` warning in the build log. Retail is 8 bytes:

```
lwz r3, TMapObjGeneral::mNormalFlushTime@sda21
blr
```

`mNormalFlushTime` already existed as a static, so the real body mirrors `getLivingTime()` directly above it. 100% match.

`MapObjInit` data goes 0.389% → 44.203%; the remainder is genuine unreconstructed content, not padding. For the whole-PR figures, see the table in the description — the ones originally quoted here were against a base that has since been superseded by the rebase.

**Codebase sweep.** I checked for the same bug elsewhere by resolving every `__vt__` owner in mario.MAP against the TU defining that class's first non-inline virtual. 624 classes resolved of 1508 vtables, and `TMapObjGeneral` is the only mismatch. An earlier version of this comment reported `JKRArchive` as a second one and said I would send it separately; that was wrong — it had already been fixed upstream in c6a3188c, which even documents the same technique:

```cpp
// NOTE: has to be defined not first to get the vtable
// to emit to the correct TU
virtual ~JKRArchive();
```

My sweep had been reading the pre-sweep header on this branch. No follow-up PR is coming, and none is needed.

KakarottoCake:
Rebased onto current `main` and refreshed the description. Dropping the WIP prefix — this is ready for review.

Three things changed since the last look:

**CI.** The `Validate symbol order` failure was not from this PR. The workflow picks its file list with a two-dot `git diff "$base" HEAD`, so a branch that is behind `main` gets every file `main` has touched added to the list and then validated in its *stale* copy — 105 units checked, 31 failed, none of them files this PR modifies. Rebasing alone took it to 10 units, all passing. Filed separately as #151 with a suggested `git merge-base` fix, since it will keep biting other PRs.

**Description.** The Results table was measured against a superseded baseline and the "Known edge case" section described the MapObjInit stand-ins, which `ca3368df` fixed. Both rewritten against the current base.

**Re-audited the PCH selection.** `main` has changed 101 source files since the original audit, and 11 of the 40 PCH units were among them, so the evidence behind the list needed redoing. Three-way controlled experiment, toggling only the PCH:

| | code | data | functions |
| - | - | - | - |
| selective (40 units) | 37.6561% | 55.1132% | 8090 |
| PCH off | 37.0842% | 50.3807% | 8063 |
| PCH global | 37.6024% | 54.7896% | 8088 |

Per-unit across the 40: 23 improve, 17 no effect, 0 regress. Across the 696 non-PCH units: none would gain, and exactly three regress under the global build — `RumbleData`, `spcinterp`, `gd-reinit-gx`, the same three the original audit excluded. So the boundary is unchanged and the selection needs no edits.

The 17 inert units are kept deliberately: seven are still at 0% code and 0% data, so there is nothing for the PCH to act on yet, and two are already at 100% data. Happy to trim them to 23 if you would rather the list carry only units with present-day evidence — it makes no difference to the build either way.

PR #133, opened 2026-08-02
GC2D/BlendPane: decompile matching TU

KakarottoCake:
## Summary
- decompile `GC2D/BlendPane` and mark the TU matching
- model the MWCC layout so `TBlendPane` has the target `0x78` size and downstream allocations match
- implement the blend update path and inline texture helper

## Verification
- `decomp-diff.py -u mario/GC2D/BlendPane`: 100% for all TU symbols/data
- `validate-symbol-order.py -u mario/GC2D/BlendPane`: PASS, 0 UNUSED symbols
- `ninja baseline` and `ninja changes_all -j 8`: pass
- clang-format and `git diff --check`: pass

This decomp TU is a test of Luna 5.6 to make sure it is ready for decompiling. It was held to full matching and validation standards. BossGesso was not touched.

### DIFF
--- a/include/GC2D/BlendPane.hpp
+++ b/include/GC2D/BlendPane.hpp
@@ -5,14 +5,11 @@
 #include <JSystem/J2D/J2DPicture.hpp>
 #include <JSystem/JUtility/JUTTexture.hpp>
 
-// Keep the data in a non-polymorphic base so MWCC places the vtable at 0x74.
-struct TBlendPaneData {
-	/* 0x0 */ f32 mBlendStep;
-	/* 0x4 */ f32 mBlendProgress;
-	/* 0x8 */ bool mBlendActive;
-};
+class TBlendPane : public TBoundPane {
+	/* 0x68 */ f32 mBlendStep;
+	/* 0x6C */ f32 mBlendProgress;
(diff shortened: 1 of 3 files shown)


include/GC2D/BlendPane.hpp
```diff
@@ -2,13 +2,34 @@
 #define GC2D_BLEND_PANE_HPP
 
 #include <GC2D/BoundPane.hpp>
+#include <JSystem/J2D/J2DPicture.hpp>
+#include <JSystem/JUtility/JUTTexture.hpp>
 
-class TBlendPane : public TBoundPane {
+// Keep the data in a non-polymorphic base so MWCC places the vtable at 0x74.
```
Mrkol: Nope, the correct way to do this is to place the fields at the top of the class. MWCC places the vtable as a field at the first definition of a virtual function, so if fields are placed before virtual functions in a class the vtable goes after them

KakarottoCake: Addressed in 8038607: removed the TBlendPaneData base and declared the blend fields directly at the top of TBlendPane before virtual update(), matching MWCC's vtable placement. The TU still matches at 100% and symbol-order validation passes.

PR #134, opened 2026-08-03
Enemy/bosstelesa: decompile Boss Telesa behavior

KakarottoCake:
## Summary
- decompile the Boss Telesa fight, roulette/slot machinery, collision actors, managers, bubbles, and all boss/bubble nerves
- restore the BossTelesa class layouts, save parameters, static tuning data, animation tables, model data, and supporting actor helpers
- reproduce the retail TU's deferred function emission order, including the interleaved BubbleLive symbols
- reconstruct the inlined `UNUSED` bodies the map requires, rather than fencing them off with `#pragma dont_inline`

## Matching

Current figures for the BossTelesa TU:

| | |
| - | - |
| fuzzy text match | 90.455% |
| exact functions | 61 / 100 |
| matched code | 8672 / 39636 bytes (21.88%) |
| matched data | 100% across `.data`, `.rodata`, `.sdata`, `.sdata2`, `.sbss` |

Note that `fuzzy_match_percent` is *lower* than it was earlier in the PR's history despite the TU being further along. Reconstructing an inlined `UNUSED` body adds a function with no counterpart on the target side, so it counts as an extra symbol and drags the average down. This TU is better judged on exact functions and matched code.

Not everything is close yet — `TBossTelesaKillSmallEnemy::checkHit` and `TBossTelesa::forceHide` are the weakest, and are genuinely unreconstructed rather than near-misses.

## Inline reconstruction

Six pragma-fenced blocks have been reduced to two (`randomReset` and `forceStopSlot`). The rest were replaced with the real call structure, using the compiler's own inlining rules rather than suppression:

- a callee is refused outright above 15 statements; otherwise a per-pass budget applies, where pass 0 (a call written directly in the function's own source) has no limit, pass 1 allows 10, pass 2 allows 7, pass 3 allows 3, and pass 4 never runs
- a call appearing as an *argument* to another call counts one pass deeper

This is what makes wrapper methods load-bearing. `TBossTelesa::tongueHitWater`, `checkMessage`, `checkSlot`, `slotStart` and `setBckAnm` all exist in the map as `UNUSED`, and routing calls through them puts the nerve-singleton constructors and `MActor` accessors at the pass depth where retail refuses to inline them.

`setBckAnm` is the clearest case. Its real body is the animation-switch block that had been open-coded in `setSpicy`; moving it there and keeping it under the pass-1 budget of 10 statements is what makes `MActor::getCurBckAnmPtr` (4 statements) land at pass 3, where MWCC gives up and emits the weak out-of-line copy the map expects from this TU. Four sizes land exactly:

| symbol | map | ours |
| - | - | - |
| `setBckAnm` | 0xd4 | 0xd4 |
| `setSpicy` | 0x19c | 0x19c |
| `TBossTelesaTongue::checkHit` | 0x23c | 0x23c |
| `getCurBckAnmPtr` (weak) | 0x1c | 0x1c |

## TU structure
- audited all 138 map entries, including all 38 `UNUSED` symbols
- all map symbols present, with linkage matching the map
- mandatory non-weak symbol order passes; what remains is compiler-controlled weak ordering and 18 best-effort `UNUSED` size warnings, i.e. inlined bodies not yet reconstructed to their retail size

## Known gaps
- 18 `UNUSED` bodies are still the wrong size, the largest being `fanfale` (map 496, ours 4), `prepareGenerate` (472 vs 12) and `calcObjCollision` (412 vs 4)
- two `#pragma dont_inline` blocks remain (`randomReset`, `forceStopSlot`); the wrapper that would replace `randomReset`'s has not been identified
- several functions differ only in stack frame size, with retail always larger; these are left unmatched rather than padded

## Verification
- `build/GMSJ01/mario.dol: OK`
- `python tools/check-changed-symbol-order.py src/Enemy/bosstelesa.cpp` — PASS
- clang-format clean
- both CI checks green


KakarottoCake:
Follow-up on the latest commit (`19063ba`) — a matching pass over the boss/bubble nerves and helpers.

## Header
- `isRollDrum()`, `rouletteFall()`, `slotFall()` changed `BOOL` → `bool`, matching the `clrlwi.` bool-test the target emits at the call sites.

## Source changes
- Rewrote the repeated *set-old-motion-blend / setBckFromIndex / setMotionBlendRatio / setAnmSound* blocks to cache `mMActor` (`MActor* oldAnmActor`, `blendActor`/`blendRatio` locals, explicit `basName` if/else) in `setSpicy`, `TNerveBossTelesaDie`, `SlotStart`, `Appear`, `PrepareSlot`.
- `genAttacker`: raw `rand() * 0.000030517578f` → `TMsRange<f32>(0.0f, 1.0f).rand()`.
- `TNerveBossTelesaDie`: unrolled `setRollSp` calls → `for` loop.
- `TNerveBubbleLive`: swapped the `unk1D0` if/else to match branch layout; `mVelocity.set(0,0,0)` → `mVelocity = JGeometry::TVec3<f32>(0,0,0)`.

## Effect
| function | before → after |
|---|---|
| TNerveBossTelesaDie | 73.9% → 94.4% |
| TNerveBossTelesaSlotStart | 79.5% → 94.9% |
| TNerveBossTelesaAppear | 85.9% → 99.5% |
| TNerveBossTelesaPrepareSlot | 83.0% → 95.7% |
| TBossTelesa::setSpicy | 87.7% → 95.7% |
| TNerveBossTelesaFallDemo | 86.6% → 90.2% |
| TBossTelesa::genAttacker | 80.8% → 88.8% |
| TNerveBubbleLive | 86.8% → 94.6% |

TU fuzzy match: 90.38% → 93.18%. Remaining residuals are MWCC stack-frame padding, register-allocation shifts, and inlining depth (e.g. `TVec3::sub`/`TMatrix34` ctor kept out-of-line in the target) — left as-is rather than faked.

KakarottoCake:
Addressed most of the review in `1ae7da3`. Notes on the tricky ones:

**Done**
- Sounds: `gpMSound->gateCheck(...)` + `MSoundSESystem::MSoundSE::startSoundActor(...)` → `SMSGetMSound()->startSoundActor(...)` with the `MSD_SE_*` enum (all 23 sites).
- Particles: added `SCENE_BTELESA_JPA_MS_BTLS_*` (+ `PARTICLE_MS_TLS_CHANGE` for `0xCD`) and switched all `SMS_LoadParticle`/`emit*` calls to the enum.
- `MsMtxSetXYZRPH` now uses the `f32` overload (drops the `(s16)(... * 182.04445f)` casts).
- Dropped the fabricated `isActorTypeOf`; the `(mActorType - base) == 1` was just `mActorType == base` — wrote the comparison inline (and `!=` for the negated cases).
- `MActor::getAnmBck` → `getCurBckAnmPtr` (matches the real `mario.MAP` name; made it the weak header inline and dropped the out-of-line def).
- `TTelesaSlot::unk198`/`unk1A8` are now `bool[3]` arrays, killing the `*(&unk198 + i)` / `u8* = &unk198 + i` fakematch.
- `rolling`/`allStopped` → `bool`; `__fabsf` → `fabsf`; `(EnumCamShakeMode)0xNN` → `CAM_SHAKE_MODE_*`; `gpMarDirector->fireStartDemoCamera` → `SMSGetMarDirector()->...`.
- Removed the `volatile f32 min/max` fakematch in `forceStopSlot`, replaced with `TMsRange<f32>`.

**Pushed back on**
- `MsWrap`: I tried `MsWrap(x, 0, 360)` for the drum-angle wrap, but the two-`if` form (`>= 360.0f` / `<= 0.0f`) is what matches 100% (`MsWrap` uses `while` + strict `<`, which generates different branches). Left it as the two `if`s.

**Still TODO (not attempted / couldn't resolve cleanly)**
- `#pragma dont_inline` blocks are still there (haven't found the real inlines they're masking).
- `checkHit`/`checkHit` UNUSED bodies are still empty stubs (sizes don't match the map).
- `getCurBckAnmPtr` body is still a guess (`return unkC;`, 8B vs map 0x1C).

**One thing to flag:** routing the sounds through the inline `SMSGetMSound()->startSoundActor` causes an inlining cascade — `TBossTelesaKillSmallEnemy::checkHit` drops 99.7% → 0% because `forceHide()` (which now carries the inline sound helper) gets inlined into it instead of staying a `bl`. A couple of other sound-bearing functions shift a few % for the same reason. I kept the sound change since it's what the cleanup wants, but wanted you to know it's not a pure win here.

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+		MtxPtr mtx = mMActor->getModel()->getBaseTRMtx();
+		MsMtxSetXYZRPH(mtx, mPosition.x, mPosition.y, mPosition.z, mRotation.x,
+		               mRotation.y, mRotation.z);
+		mMActor->getModel()->setBaseScale(mScaling);
+	}
+}
+
+void TBubble::setDeadAnm() { setBckAnm(9); }
+
+MtxPtr TBubble::getTakingMtx() { return mMActor->unk4->unk20; }
```
Mrkol: Use the inline getters.

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+		mHeldObject = enemy;
+		JGeometry::TVec3<f32> velocity;
+		velocity.set(0.0f, 1.0f, -1.0f);
+		enemy->mVelocity = velocity;
+		enemy->onLiveFlag(LIVE_FLAG_AIRBORNE);
+		unk198 = enemy;
+	}
+}
+
+BOOL TNerveBubbleLive::execute(TSpineBase<TLiveActor>* spine) const
```
Mrkol, line 330: Nerve definitions/declarations should use the macros

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+	TEnemyManager::perform(flags, gfx);
+}
+
+BOOL TBossTelesaBody::receiveMessage(THitActor* sender, u32 message)
+{
+	TBossTelesa* boss = unk68;
+
+	if (message == HIT_MESSAGE_TRAMPLE) {
+		bool isMario;
+		if (sender->mActorType == (ACTOR_TYPE_PLAYER | 1))
```
Mrkol: ACTOR_TYPE_PLAYER etc macros should not be used like this -- just hardcode the actor type. This part is to be refactored later on.

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+		else
+			basName = basTable[4];
+		boss->setAnmSound(basName);
+		boss->mMActor->setBtpFromIndex(2);
+	}
+
+	if (boss->checkCurAnmEnd(0)) {
+		boss->onHitFlag(HIT_FLAG_NO_COLLISION);
+		boss->unk16C->onHitFlag(HIT_FLAG_NO_COLLISION);
+		boss->unk170->onHitFlag(HIT_FLAG_NO_COLLISION);
```
Mrkol: this is offAllCollisions. Check the entire code for other other functions from the same file being inlined

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+		    == &TNerveBossTelesaPrepareSlot::theNerve())
+			boss->mSpine->pushNerve(&TNerveBossTelesaFreeze::theNerve());
+	}
+
+	return TRUE;
+}
+
+void TBossTelesaBody::checkHit() { }
+
+void TBossTelesaTongue::checkHit() { }
```
Mrkol: these need to be reconstructed

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+		bubble->mPosition.y += 10.0f;
+		bubble->onLiveFlag(LIVE_FLAG_AIRBORNE);
+
+		TMsRange<f32> itemRange(0.0f, 1.0f);
+		f32 itemRoll = itemRange.rand();
+		if (itemRoll < mItemGenRate) {
+			bubble->unk198    = nullptr;
+			TMapObjBase* item = gpItemManager->makeObjAppear(
+			    bubble->mPosition.x, bubble->mPosition.y, bubble->mPosition.z,
+			    0x20000008, true);
```
Mrkol: This place should be the appendItem inline -- reconstruct it based on size. It's symmetric to appendEnemy it seems

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+				if (i == 0) {
+					velocity.x
+					    = normalizedDir.x * fruitSpeedRange.rand() * 2.0f;
+					velocity.z
+					    = normalizedDir.z * fruitSpeedRange.rand() * 2.0f;
+				}
+
+				((TMapObjBase*)unk2F8[i])->mVelocity = velocity;
+				((TMapObjBase*)unk2F8[i])->offLiveFlag(LIVE_FLAG_UNK10);
+				((TMapObjBase*)unk2F8[i])->mRotation.set(0.0f, 90.0f, 0.0f);
```
Mrkol: Here and in a bunch of places casts are used for zero reason -- use the exact type in the field instead (forward-declare it in the header so as not to include a bunch of stuff)

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+			dir.y = 10.0f;
+
+			Vec velocity;
+			TMsRange<f32> itemSpeedRange(0.8f, 3.5f);
+			f32 speed  = params->mSL1stBubbleSp.get();
+			velocity.x = dir.x * speed * itemSpeedRange.rand();
+			velocity.y = dir.y;
+			velocity.z = dir.z * speed * itemSpeedRange.rand();
+
+			TMapObjBase* item = gpItemManager->makeObjAppeared(0x2000000E);
```
Mrkol: Better to cast to TItem* here, that makes more sense as code that humans would've wrote

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+			velocity.y = dir.y * (2.0f + enemySpeedRange.rand());
+			velocity.z = dir.z * speed * enemySpeedRange.rand();
+
+			enemy->mPosition.x = rootMtx[0][3];
+			enemy->mPosition.y = rootMtx[1][3] - 250.0f;
+			enemy->mPosition.z = rootMtx[2][3];
+			enemy->mVelocity   = velocity;
+			enemy->mPosition.y += 10.0f;
+			enemy->onLiveFlag(LIVE_FLAG_AIRBORNE);
+			PSMTXCopy(rootMtx, enemy->mMActor->unk4->unk20);
```
Mrkol: Here and everywhere else, the wrapper macros should be used: MTXCopy instead of PSMTXCopy

src/Enemy/bosstelesa.cpp
```diff
@@ -1 +1,3124 @@
...
+}
+
+void TBossTelesa::fruitCollisionOn() { unk370 = 0; }
+
+bool TBossTelesa::checkSlot() { return unk184 != nullptr; }
+
+bool TBossTelesa::checkAllItemDead()
+{
+	for (int i = 0; i < unk274; ++i)
+		if (!(unk1AC[i]->mLiveFlag & LIVE_FLAG_DEAD))
```
Mrkol: checkLiveFlag inline should be used here and everywhere else -- validate that appropriate inlines are used

KakarottoCake:
Pushed two more commits addressing feedback:

**`d78a664` — use MActor motion-blend accessors**
Collapsed 42 copies of the open-coded motion-blend boilerplate (null-checking `unkC`, pulling `unk24`, calling `setOldMotionBlendAnmPtr`/`setMotionBlendRatio` by hand) into the existing `MActor` wrapper calls (`getBckAnm()`, `setBckOldMotionBlendAnmPtr()`, `setMotionBlendRatioForBck()`). Net -183 lines, no regressions (had to cache `boss->mMActor` in a local before the two calls to keep codegen matching register allocation on the double-indirect case).

**`fc9fd83` — reconstruct the inlined checkHit bodies**
`TBossTelesaBody::checkHit` and `TBossTelesaTongue::checkHit` are `UNUSED` in mario.MAP (fully inlined into `moveObject`). Moved the open-coded collision loops out of `moveObject` and back into these methods, matching the map. Also swapped the hand-rolled `isEnemy` bool + branch for `checkActorType(ACTOR_TYPE_ENEMY)`, and the player-type comparison for `isActorType(0x80000001)`, per the actor-type inline feedback.

Still open from the review: the 6 remaining `#pragma dont_inline` fakematches, and ~20 other `UNUSED` bodies that still need reconstructing to their map sizes (fanfale, checkMessage, prepareGenerate, calcObjCollision, tongueHitWater, appendItem, etc.). Also flagging two spots where I think the review note and the ROM disagree — will follow up in a separate comment/reply on those threads rather than silently going against them.

KakarottoCake:
_Edited: this predated the rebase onto `main` and the `d2c364b9` fix. The commit SHAs below were rewritten by the rebase, and the open question at the end has since been answered — corrected inline so nobody spends time on it._

Status update — five commits.

**match `TBubble::kill`**
`onLiveFlag(LIVE_FLAG_UNK20000)` belongs inside the nerve check, not after it. The ROM's early-exit branch lands on the `0x40` flag block and skips the `0x20000` one, so when the current nerve is already `TNerveSmallEnemyDie` only `0x40` is set. 100%.

**`tongueHitWater` and `checkMessage` wrappers**
Both collision-actor `receiveMessage` overrides forward to a `TBossTelesa` method instead of touching the spine directly. Both are UNUSED in the map, so they inline straight back and the emitted sizes are unchanged.

The extra layer is what fixes the nerve singletons. Reaching `pushNerve` directly put `TNerveBase<TLiveActor>::TNerveBase()` at inline pass 3, where its 2 statements fit the budget of 3 and it got inlined; the ROM has a `bl`. The wrapper pushes it to pass 4, which never runs. Both overrides now match 100%.

**reconstruct `checkSlot`**
The tail of the SlotStart nerve is a bool-returning method (UNUSED, 0xd4), not an inline early return. An inlined bool-returning function whose result is then tested is what emits the ROM's `li r0,1 / b / li r0,0 / clrlwi. / beq`; we were branching directly.

**reconstruct `slotStart`, drop four `dont_inline` pragmas**
`moveStart` is 19 statements, so at pass 0 — which has no limit — it expanded, and the pragma existed only to stop that. Routing through `TBossTelesa::slotStart()` puts it at pass 1 where 19 > 10, so it stays out-of-line on its own. The pragmas around `isRollDrum`, `getSlotResult` and `forceAllItemKill` turned out to do nothing at all — byte-identical with or without — and were removed.

**`getCurBckAnmPtr` naming**
Per your answer: the null-checked `unkC->unk24` body returning `J3DAnmTransformKey*` takes the real name, and the 8-byte `return unkC;` getter becomes `getUnkC()`. The map confirms it — one entry, 28 bytes, only in this TU.

---

**The open question in the original version of this comment is now resolved, in `d2c364b9`.** I had asked which method causes `getCurBckAnmPtr__6MActorFv` to be emitted, and reported that I had tried `setBckAnm` and reverted it after it made `TNerveBossTelesaSpit::execute` worse.

`setBckAnm` was in fact the right answer; I had it at the wrong depth. A compiler inline trace gives `getCurBckAnmPtr` as **4 statements**, so it is refused only at pass 3 (budget 3), never at pass 2 (budget 7). The chain that produces the emission is:

```
TBossTelesaTongue::checkHit (source)
  -> setSpicy                     pass 0  inlined
  -> setBckAnm                    pass 1  inlined - must be <= 10 statements
  -> setBckOldMotionBlendAnmPtr   pass 2
  -> getCurBckAnmPtr (nested arg) pass 3  REFUSED -> emitted
```

My first attempt routed it from the Spit nerve instead, which is one layer too shallow, and `setBckAnm` was 15 statements — small enough to inline at pass 0 but refused at pass 1, so the chain died one link short and nothing was emitted. Collapsing one `if/else` in it to a ternary brought it under 10. `setBckAnm`'s real body is the animation-switch block that had been open-coded in `setSpicy`.

Four sizes now land exactly on the map — `setBckAnm` 0xd4, `setSpicy` 0x19c, `TBossTelesaTongue::checkHit` 0x23c (was 0xec), and `getCurBckAnmPtr` 0x1c (was absent) — and it is emitted at exactly the map's position, immediately after `checkHit`. `check-changed-symbol-order.py` goes from FAIL to PASS. No question outstanding here.

For current figures see the description; the ones originally quoted in this comment were against a base the rebase has since superseded.

src/Enemy/bosstelesa.cpp
KakarottoCake: Now uses the accessors: \`mMActor->getModel()->getBaseTRMtx()\` instead of \`mMActor->unk4->unk20\`.

src/Enemy/bosstelesa.cpp
KakarottoCake: Hardcoded — \`(ACTOR_TYPE_PLAYER | 1)\` is now \`0x80000001\`, and the mask test is now \`(actor->mActorType & 0xFFFF0000) != 0x40000000\`.

src/Enemy/bosstelesa.cpp
KakarottoCake: Both reconstructed. \`TBossTelesaTongue::checkHit\` is now 0x23c and \`TBossTelesaBody::checkHit\` 0xa0, matching the map exactly.

src/Enemy/bosstelesa.cpp
KakarottoCake: Reconstructed \`TBubble::appendItem()\` from the block that was open-coded at the generation site, mirroring \`appendEnemy\`. Compiles to 0xd8, exactly the map size, and the site now just calls \`bubble->appendItem()\`.

src/Enemy/bosstelesa.cpp
KakarottoCake: Changed the field types — \`unk2A8\` and \`unk2F8\` are \`TMapObjBase*\` now, which removed twelve casts at the use sites.

src/Enemy/bosstelesa.cpp
KakarottoCake: Hoisted: \`TItem* item = (TItem*)gpItemManager->makeObjAppeared(0x2000000E);\`, so the later \`((TItem*)item)->killByTimer\` cast is gone.

src/Enemy/bosstelesa.cpp
KakarottoCake: Switched to the wrapper macros throughout — MTXCopy, MTXConcat and MTXMultVec. No PSMTX* calls remain in the file.

src/Enemy/bosstelesa.cpp
KakarottoCake: Now uses \`checkLiveFlag()\` everywhere a flag is tested. The only remaining raw \`mLiveFlag\` uses are the \`&actor->mLiveFlag\` address-of ones in the fruit code, which cannot go through the inline.

src/Enemy/bosstelesa.cpp
KakarottoCake: Good catch, and it went further than the call site. `offAllCollision`/`onAllCollision` were touching a fourth collision actor (`unk174`) — with it they compile to 0x40, while the map wants 0x30 for both, which is exactly one `onHitFlag` too many. Dropped it, and both now land on 0x30.

I then swept the file for the same pattern and found three sites with the identical three-call sequence, at the Damage, Die and Hide nerves; all three now call `boss->offAllCollision()`. Byte-identical output, so they were indeed the inline.

src/Enemy/bosstelesa.cpp
KakarottoCake: Not done — I think this one has to stay hand-written, and I would rather flag it than quietly skip it.

`DEFINE_NERVE` emits `theNerve()` and `execute()` adjacent. The map has them separated by two other functions:

```
002d2c38 0003e8  execute__16TNerveBubbleLiveCFP24TSpineBase<10TLiveActor>
002d3020 000214  appendEnemy__7TBubbleFv
UNUSED   0000d8  appendItem__7TBubbleFv
UNUSED   000064  theNerve__16TNerveBubbleLiveFv
002d3234 00005c  __dt__16TNerveBubbleLiveFv
```

Since .text emission is reverse definition order, that means the retail source defined `theNerve`, then `appendItem`, then `appendEnemy`, then `execute` — so the two halves cannot be adjacent.

I tried it anyway to be sure: converting to `DEFINE_NERVE` makes `check-changed-symbol-order.py` fail with `Non-weak symbol ORDER differs from the map`. Reverted. Every other nerve in the file does use the macro — this is the only one that cannot. Happy to add a comment above it explaining why if you would like.

src/Enemy/bosseel.cpp
```diff
@@ -1917,7 +1917,7 @@ void TBossEel::setBckAnm(int index)
 	mPreviousBckIndex = getMActor()->getCurAnmIdx(0);
 	mCurrentBckIndex  = index;
 	mBckBlendRatio    = 1.0f;
-	getMActor()->setBckOldMotionBlendAnmPtr(getMActor()->getBckAnm());
+	getMActor()->setBckOldMotionBlendAnmPtr(getMActor()->getCurBckAnmPtr());
```
Mrkol: Reminder that getCurBckAnmPtr is 0x1C and so shouldn't replace getBckAnm -- it's a different method it seems

Mrkol (changes requested):

src/Enemy/bosseel.cpp
KakarottoCake: Confirmed — restored `getBckAnm` and put `bosseel.cpp` and `CameraBck.cpp` back on it; `bosstelesa.cpp` keeps `getCurBckAnmPtr`, which is still emitted weak at exactly 0x1c. `getBckAnm` is emitted in neither object, which is why it is absent from the map. No match change: 8131/12881 before and after.

src/Enemy/bosstelesa.cpp
KakarottoCake: Added the note above theNerve in 42ea2c3b. Merged origin/main as well, no conflicts in this TU.

PR #135, opened 2026-08-03
GMSP01: decompile video height getters

999sian:
In GMSP01 the three video *height* getters take the TV format from the caller (`__FUl`) instead of querying VI themselves (`__Fv` in GMSJ01). They were undecompiled, which was two of the six undefined symbols at PAL link time.

`System/Resolution.cpp` is now 12/12 matching for GMSP01 (`SMSGetGCLogoVideoHeight` 0x20, `SMSGetTitleVideoHeight` 0x40, `SMSGetGameVideoHeight` 0x40). Structure is unchanged from the JPN source; only the format source differs, and the GCLogo -> Title -> Game chain still collapses to a single `bl SMSGetGameVideoHeight__FUl` via one-level inlining, as in the retail binary.

Call sites updated so GMSP01 still compiles. Reusing the format each function already queries also improves `RenderModeObj` 74.23% -> 78.86% (`SMSSetupGCLogoRenderMode` 79.58% -> 87.92%, `SMSSetupTitleRenderMode` 81.91% -> 88.21%).

GMSJ01 is unaffected: `Resolution.cpp`, `RenderModeObj.cpp` and `MarDirectorSetupObjects.cpp` all compile to byte-identical objects before and after (checked by compiling both revisions with `-DVERSION_GMSJ01` and comparing).

Verification (GMSP01): `validate-symbol-order.py` passes presence/order/linkage on all three TUs against `marioEU.MAP`; symbol order including the two UNUSED entries matches the map exactly.

Open question: this is the first use of `-DVERSION_GMSP01`, which `configure.py` has always emitted but nothing consumed. Happy to reshape it if you'd prefer a different mechanism for version divergence.

Not addressed:
- Retail PAL derives the format once per function and overrides it via `OSGetConsoleType()` / `OSGetEuRgb60Mode()`. That isn't implemented, so those functions stay non-matching.
- `MarDirectorSetupObjects` goes 96.89% -> 96.85%; the target inlines the call there and we can't inline cross-TU.
- PAL still doesn't link: `OSGetEuRgb60Mode`, `OSSetEuRgb60Mode`, `OSGetLanguage` (OSRtc) remain undefined.
- Unrelated, but `marioEU.MAP` lists `SMSGetDisplayWidthMax`/`SMSGetDisplayHeightMax` as UNUSED at 0x74; the current `return 720` / `return 574` bodies compile to 0x8, so PAL's are probably format-dependent too.

### DIFF
--- a/include/System/Resolution.hpp
+++ b/include/System/Resolution.hpp
@@ -3,7 +3,11 @@
 
 #include <dolphin/types.h>
 
+#ifdef VERSION_GMSP01
+u16 SMSGetGCLogoVideoHeight(u32 format);
+#else
 u16 SMSGetGCLogoVideoHeight();
+#endif
 u16 SMSGetGCLogoVideoWidth();
 u16 SMSGetGCLogoRenderHeight();
 u16 SMSGetGCLogoRenderWidth();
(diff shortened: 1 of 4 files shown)


Mrkol:
The PAL version is not hooked up to CI so I'd like to see that before accepting any PRs targeting it

999sian:
? What do you want me to do

PR #136, opened 2026-08-09
SelectMenu matching and naming.

CaptainProton42:
There are still tons of issues with this but I think I've stared at this for long enough and am reaching the end of my abilities.

I've done the naming mostly by running the game with Dolphin Memory Engine attached.

I've also moved TSelectShineManager to its own file to mirror the structure of the TUs.


include/GC2D/SelectMenu.hpp
```diff
@@ -1,55 +1,193 @@
-#ifndef GC2D_SELECT_MENU_HPP
-#define GC2D_SELECT_MENU_HPP
+#ifndef GC2D_SELECT_MENU_2_HPP
+#define GC2D_SELECT_MENU_2_HPP
```
Mrkol: Why 2 when there's no 2 in the file name?

PR #138, opened 2026-08-19
feature(decomp): match and link MSL ansi_fp

mattsumi:
Matches `ansi_fp.c` fully and flips it to `Matching`. The DOL sha1-verifies with the unit linked.

## What was needed

`__num2dec` already matched 100% per-symbol, but linking the unit broke the DOL hash: the `.sdata2` literal pool order was wrong. The cause is the UNUSED `__dec2num` — under `-inline deferred` it is codegen'd first and pools the `0.0`/`1.0`/int-conversion-magic literals (`@268`/`@270`/`@272`) before `__num2dec` adds `0.1`/`10.0` (`@362`/`@363`). Reconstructing a `__dec2num` body that touches exactly those constants in that order restores the original pool layout. objdiff's per-symbol view cannot see this class of mismatch; only the DOL check catches it.

The reconstruction is a plausible inverse of `__num2dec` (digit-chunk accumulation via `digit_values`, power-of-ten scaling via `bit_values`), written from the binary, the linker map, and the emitted literal pool only — no external MSL sources were consulted. It is smaller than the map's record (0x13c vs 0x2a4, noted with a TODO); since the function is deadstripped, only its literal-pool side effect and symbol presence affect the link.

## Verification

- `ninja` — `build/GMSJ01/mario.dol: OK` with ansi_fp linked (SDK 147/149 linked)
- `tools/validate-symbol-order.py` — PASS (one non-failing UNUSED size warning, noted in-source)
- all symbols 100% in objdiff, `.sdata2` byte order identical to target

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -576,7 +576,7 @@ config.libs = [
             Object(Matching, "PowerPC_EABI_Support/Msl/MSL_C/MSL_Common/ansi_files.c"),
             Object(Matching, "PowerPC_EABI_Support/Msl/MSL_C/MSL_Common/abort_exit.c"),
             Object(Matching, "PowerPC_EABI_Support/Msl/MSL_C/MSL_Common/errno.c"),
-            Object(NonMatching, "PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/ansi_fp.c"),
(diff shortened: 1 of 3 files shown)


Mrkol:
Didn't you see the prohibition on clanker matching in MSL/SDK? It's just pulling out leaked sources from it's memory.

PR #139, opened 2026-08-19
rework(decomp): improve powf structural match in exponentialsf

mattsumi:
Takes `powf` from 96.0% to 98.1% fuzzy match. Every opcode now matches; the residue is FP register renumbering plus a 0x10 stack frame delta, documented with TODOs in-source.

## Changes

- The x < 0 path keeps `iy = (int)y` **and** an inline `(float)(int)y` cast: the target materializes two separate conversion temps from one `fctiwz` at the def point, which only happens when the two uses are distinct expressions.
- The fabricated x/y classification switches are actually MSL's `fpclassify` — using the existing `__fpclassifyf` header inline reproduces the low-frame temp the target shows (a named local copy places it wrongly).
- `__log2f`: the lo-bits mask is hoisted into a register temp (`lo_mask`) before the halfword test, and the `bits` local is dropped in favor of direct `*(int*)&x` reads, which fixes the GPR assignment around the three inline sites.
- `__exp2f`: the final multiply reuses `p` instead of a fresh local, removing one stack object without refusing the target's unfused `fmuls`.

`expf` (97.2%) shares the remaining FP-renumbering puzzle: the target keeps its long-lived constants in f6-f9 while we allocate f0-f4, so the Horner chain temps start at the wrong watermark. Source-order and const-local/inline-literal permutations were tried and do not flip the allocation; noted in-source.

## Verification

- unit compiles clean; `powf` 98.1% / `expf` 97.2%, zero opcode-level diffs in `powf`
- all rodata/sdata2 symbols 100%
- no regressions elsewhere (`ninja changes_all` clean)

### DIFF
--- a/src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/exponentialsf.c
+++ b/src/PowerPC_EABI_Support/Msl/MSL_C/MSL_Common_Embedded/Math/Single_precision/exponentialsf.c
@@ -60,6 +60,11 @@ extern const float __two_to_log2e_m1_tI[178];
 #define V_EXPF_MAX (88.72284f)
 #define V_EXPF_MIN (-87.33655f)
 
+// TODO: nonmatching, instruction stream is exact but FP registers are
+// renumbered: the target keeps the long-lived values (constants, pow2,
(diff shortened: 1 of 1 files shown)


Mrkol:
Same as the other one -- don't touch MSL/SDK with dirty clanker fingers.

PR #143, opened 2026-08-19
feature(decomp): match AnimalManager

mattsumi:
Matches Animal/AnimalManager, with matching improvements to AnimalNerve and Butterfly (remaining residues documented with TODO notes).

## Verification
- full build from upstream main + this change: `build/GMSJ01/mario.dol: OK`

### DIFF
--- a/src/Animal/AnimalManager.cpp
+++ b/src/Animal/AnimalManager.cpp
@@ -21,25 +21,36 @@ TAnimalManagerBase::TAnimalManagerBase(const char* name)
 	mAnimalSave     = nullptr;
 }
 
+void TAnimalManagerBase::loadSaveParams_(const char* prm)
+{
+	mAnimalSave     = new TAnimalSaveIndividual(prm);
+	mViewClipNear   = mAnimalSave->mSLViewClipNear.get();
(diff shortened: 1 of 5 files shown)


src/Animal/AnimalManager.cpp
```diff
@@ -21,43 +21,54 @@ TAnimalManagerBase::TAnimalManagerBase(const char* name)
...
+	loadSaveParams_("/Animal/mew.prm");
 }
 
 void TMewManager::loadAfter()
 {
 	TAnimalManagerBase::loadAfter();
-	MSoundSESystem::MSRandPlay::createRandPlayVec(MSD_SE_OBJ_KAMOME_SOLO,
-	                                              mObjNum);
+	// TODO: fabricated - this unused local is required to match the stack
+	// frame size (0x28); TAnimalBase::loadAfter has the same extra 0x10
```
Mrkol: This is a fakematch, we don't do this around here.

Mrkol:
One PR per TU please. Also, no dumb clanker comments and "TODO: match  this" crap -- we can see in objdiff that it's not matching, no need for a todo.

PR #146, opened 2026-08-19
feature(decomp): match MapObjFloat and MapObjWater

mattsumi:
Matches MoveBG/MapObjFloat and MapObjWater, with matching improvements to the other changed MoveBG TUs. The `MapObjManager.hpp` change also appears identically in the companion core PR; whichever merges first, the other rebases clean.

## Verification
- full build from upstream main + this change: `build/GMSJ01/mario.dol: OK`


Mrkol:
One PR per TU. And stop it with the quantity. We need quality, not a river of slop.

PR #150, opened 2026-08-20
feature(decomp): match GC2D/HelpActor

mattsumi:
Matches GC2D/HelpActor and links it. One source file plus the configure.py flip.

Verified: full build from main, mario.dol sha1 OK, validate-symbol-order passes.


RevoSucks:
I highly advise you to take a step back and hold off of any more PRs.

The contributor(s) of SMS value "correctness" more than forced matches using stack to pad the stack. The compiler typically leaves stack as a result of inlines/getters/setters/etc and these are typically 'fake' matches. They do not add value to the repository.

mattsumi:
👍 keeping this in mind.

PR #157, opened 2026-09-03
Decompile Boss Pakkun

KakarottoCake:
`src/Enemy/bosspakkun.cpp` was empty on main. This fills it in, along with
`include/Enemy/BossPakkun.hpp`.

82 of 126 functions match. 293 of 296 data objects match; `.bss` and `.sdata2`
are exact and `.rodata` is 7 bytes short. The unit stays NonMatching.

The TU needs `M3DUtil/InfectiousStrings.hpp`. Without it we were 224 bytes short
in `.rodata` ahead of the string table, which shifted every pool reference in
`TBossPakkunParams::TBossPakkunParams` and `TBossPakkunManager::load`. Both match
exactly once it is included.

Known gaps:

- 13 functions differ only in stack frame size. Left alone.
- `TNerveBPDie::execute` (80.2%) is the last structural one. Retail repeats
  `modBgm(0, 1)` in both arms of a branch and the source factors it into a single
  call. I did not duplicate an identical call to gain the score.
- Three UNUSED bodies are short of their map sizes: `ignoreWaterCheck` (0x4 vs
  0xf0), `initJParticle` (0x274 vs 0x280), `vomitFinished` (0x40 vs 0x3c).
- `changeBck` spells the previous-index transitions as separate `if` branches
  instead of a `switch` with fall-through cases. Retail emits its own parameter
  call per branch, so the fall-through form does not reproduce it, but the `if`
  form does repeat one line three times. Happy to change it back if you would
  rather have the `switch`.


src/Enemy/bosspakkun.cpp
```diff
@@ -1 +1,2047 @@
...
+	    ->getChildren()
+	    .push_back(this);
+}
+
+void TBPPolDrop::drop()
+{
+	unk80 = 2;
+	unk6C.zero();
+	unk7C->setBck("pollut_ball_stamp");
+	gpMarioParticleManager->emit(0x52, &mPosition, 0, nullptr);
```
Mrkol: Give name to all particles pls as we always do

src/Enemy/bosspakkun.cpp
```diff
@@ -1 +1,2047 @@
...
+	unk14->setBckFromIndex(-1);
+	unk18->setBckFromIndex(-1);
+}
+
+void TBPVomit::perform(u32 flags, JDrama::TGraphics* graphics)
+{
+	if (unk14->getCurAnmIdx(ANM_TYPE_BCK) < 0)
+		return;
+
+	u32 calcAnim = flags & CUE_CALC_ANIM;
```
Mrkol: Don't do this, just do `flags & CUE_CALC_ANIM` twice -- CSE will pull it out to a single calculation on it's own

src/Enemy/bosspakkun.cpp
```diff
@@ -1 +1,2047 @@
...
+				vanish();
+				return;
+			}
+
+			f32 angle = 360.0f - static_cast<s32>(unk94) % 360;
+			f32 radius
+			    = unk94
+			      * mOwner->getBossPakkunParams()->mSLTornadoRollSpeed.get();
+			JGeometry::TVec3<f32> direction;
+			direction = unk70;
```
Mrkol: Why? Just assign when declaring the variable

src/Enemy/bosspakkun.cpp
```diff
@@ -1 +1,2047 @@
...
+			direction.sub(unk7C);
```
Mrkol: It's usually `-=` rather than just `sub`

src/Enemy/bosspakkun.cpp
```diff
@@ -1 +1,2047 @@
...
+		emitter = gpMarioParticleManager->emitAndBindToMtxPtr(0x164, mtx, 1,
+		                                                      (u8*)this + 2);
+		if (emitter)
+			emitter->setGlobalScale(mScaling);
+	}
+
+	if (flags & CUE_CALC_ANIM) {
+		JGeometry::TVec3<f32> toMario;
+		toMario = mPosition;
+		toMario.sub(*gpMarioPos);
```
Mrkol: use `-=` and `SMS_GetMarioPos`

src/Enemy/bosspakkun.cpp
```diff
@@ -1 +1,2047 @@
...
+	rotation[0][2] = 0.0f;
+	rotation[0][3] = 0.0f;
+	rotation[1][0] = 0.0f;
+	rotation[1][1] = c;
+	rotation[1][2] = -s;
+	rotation[1][3] = 0.0f;
+	rotation[2][0] = 0.0f;
+	rotation[2][1] = s;
+	rotation[2][2] = c;
+	rotation[2][3] = 0.0f;
```
Mrkol: MsMtxSetRotX

Mrkol (changes requested):

src/Enemy/bosspakkun.cpp
KakarottoCake: Done. 0x52 is ms_osendan.jpa, added PARTICLE_MS_OSENDAN, plus SCENE_BOSSPAKKUN_JPA_MS_BOPA_* for the 12 boss loads. All emit and load sites use the names.

src/Enemy/bosspakkun.cpp
KakarottoCake: Done. Dropped the calcAnim local in both perform functions, tests flags directly.

src/Enemy/bosspakkun.cpp
KakarottoCake: Done. Declares and inits on one line now.

src/Enemy/bosspakkun.cpp
KakarottoCake: Done. Uses -= now.

src/Enemy/bosspakkun.cpp
KakarottoCake: Done. Uses SMS_GetMarioPos with -=.

src/Enemy/bosspakkun.cpp
KakarottoCake: Done. Uses MsMtxSetRotX now.

PR #158, opened 2026-09-04
docs: update macOS build instructions

MagiHotline:
### What
A very small PR that changes the README instructions for arm64 macOS.

### Reasons

The macOS build instruction were outdated since the developers of Wine Crossover officially removed their binary from the brew packages since it's no longer maintained. In this PR I take for granted that every macOS user has Rosetta 2 (s a macOS component that automatically translates Intel-based) already installed. 

### Testing Done
Verified from a clean clone on macOS:
1. Configured via `python3 configure.py [args]`
2. Built successfully with `ninja` (up to checksum validation)

### Problems encountered

During building, `wibo` failed because it is trying to execute `sjiswrap.exe`, which uses Win32 API functions that wibo does not implement. Since `sjiswrap.exe` is only needed on systems where the compiler cannot handle Shift-JIS source files properly under Wine, I decided to bypass it via: 

```bash
python3 configure.py --wrapper /usr/local/bin/wibo --sjiswrap build/compilers/GC/1.2.5/mwcceppc.exe
```
After this, it got fixed. But I'm not really sure to add this in the README too.


Mrkol:
Lets just update the readme to latest dtk-template version rather than have our own wording/instructions
https://github.com/encounter/dtk-template/blob/main/README.example.md#macos

MagiHotline:
Does the NInja Build pull the right wibo if MacOs system is detected?

Mrkol:
> Does the NInja Build pull the right wibo if MacOs system is detected?

IDK, maybe we need to update other parts of dtk-template too so it fetches the right thing

MagiHotline:
I bumped wibo version to where macOs machines could get automatically downloaded the right release. I also needed to add a config flag so that wibo executes `mwccppc` directly without sjiwrap since it would have panicked otherwise

Mrkol:
@MagiHotline rebase please

MagiHotline:
rebbbbased

Mrkol:
I updated us to latest dtk-template changes so it should be good now

PR #161, opened 2026-09-15
SMS decomp: current harness progress

fjooord:
## Summary

Harness snapshot of accumulated SMS work, force-updated on 2026-09-16 after a source-fidelity audit, lint cleanup, and re-land. Head `3a5c585b`, which merges upstream `9c2f2f3b`.

## What changed since the last push

- Merged upstream through `9c2f2f3b` (35 commits, #162 to #187 plus the toolchain bump to dtk v1.8.4, objdiff v3.8.1, wibo 1.1.0). In the final merge six files that conflicted were taken whole from upstream: Menu.cpp, GCLogoDir.cpp, MarDirectorEvent.cpp, MarNameRefGen_BossEnemy.cpp, MarNameRefGen_Enemy.cpp, MarioGamePad.cpp.
- 820 further worker integrations since the previous snapshot `aea424cd`.
- A source-fidelity audit found accepted integrations that reached their score with constructs no original developer would have written: hand-mangled `extern "C"` statics, hand-written vtable arrays, discarded literals, unused padding locals, guard-override macros, unmarked pragmas. Those rules are now enforced by the harness lint gate.
- 104 integrations that fail the new lints were handled: 62 reverted (each `revert(lint):` commit names the failed rules and the target to requeue; git notes mark the original commits), 42 already superseded by the upstream merge. One transiently flagged definition (`sFogOffColor`) was restored because a later clean integration uses it.
- 31 `reland(lint):` commits then restored the reverted data the way the original code would have spelled it: class statics declared in the owning headers (`include/Enemy`, `include/MoveBG`) and defined in map order, PCH string includes instead of hand copies, an inline destructor in `FeetInv.hpp`, and the `FifoSetFog` body. Fabricated header classes are marked as such.

## Validation

- Full Linux rebuild in a Daytona sandbox from the sandbox image: all 740 `all_source` objects compile, `progress` and `report.json` generate.
- Versus the pre-cleanup report: matched code 40.92% to 40.73%, matched functions 8432 to 8395 of 12881, matched data 62.55% to 59.65%. Every change is inside the reverted units; nothing outside them moved. 17 sections and 1 function came back through the re-land; 63 sections and 50 functions remain below their pre-cleanup score and are the requeue set.
- The cumulative delta against upstream scans at zero lint errors (49 review-tier warnings). Rebuilt in Daytona after the merge: 739 objects, 0 failures.
- Symbol-order validation could not run: `orig/GMSJ01/files/mario.MAP` is not in the image or the checkout.

## Pending

- Requeue the reverted targets through the harness board at the next epoch.
- The two formatting and CI commits previously on this branch (`58874c3c`, `f37da262`) were dropped by the force update; their CI content is upstream via #162.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fjooord:
Superseded by #189, a single squashed commit with an identical tree on top of upstream 9c2f2f3b. This branch keeps the full harness history (per-target revert(lint) and reland(lint) commits with git notes) for provenance.

PR #162, opened 2026-09-15
CI: speed up formatting and check symbol regressions against the base

fjooord:
## 1. Check All Formatting in One Container

Run clang-format 21 across all tracked source files in one Docker container. No automatic edits or custom formatting helper.

- Previous CI: **9m23s**, starting a container per file.
- Local check: **15.7s for 1,475 files**, with the image cached. Remote timing pending.
- Includes the whitespace-only fix in `include/JSystem/JAudio/JALibrary/JALList.hpp`, the only file failing formatting. The full-tree check now passes locally.

## 2. Fail on New Symbol Defects

#161 compiled and generated its report, then failed symbol-map validation. Compare against a separately built upstream base: inherited defects become warnings; new missing symbols, ordering errors, or incorrect linkage fail CI.

**Previously empty objects get no exemptions.** If the base object has no function symbols, its first implementation must pass strict validation.

Compiler errors and missing validation inputs still fail. Standalone validation stays strict. This does not implement missing functions or repair existing symbol defects.

**Validation:** 15 local checks pass, including empty-baseline behavior. No test files or test steps are added to the PR. Game implementation changes stay in #161.


.github/workflows/build.yml
```diff
@@ -89,7 +94,27 @@ jobs:
...
+          exit 0
+        fi
+        baseline="$RUNNER_TEMP/symbol-baseline"
+        git worktree add --detach "$baseline" "$BASE"
+        mkdir -p "$baseline/orig"
+        cp -a /orig/. "$baseline/orig/"
+        cd "$baseline"
+        python configure.py --map --version ${{ matrix.version }} \
+          --binutils /binutils --compilers /compilers
+        ninja all_source
```
Mrkol, line 111: Empty files which no one started decompiling yet should get a "no errors" baseline so that PRs that implement them are forced to pass the validation.

tools/test_symbol_validation.py
```diff
@@ -0,0 +1,118 @@
...
+        self.assertEqual(set(formatting.select_files(self.base)),
+                         {"changed file.cpp", "unchanged.hpp", "deleted.cpp"})
+
+    def test_invalid_base_fails(self):
+        with self.assertRaises(subprocess.CalledProcessError):
+            formatting.select_files("nonexistent")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
Mrkol: I don't think we want CI tests in master, it's on the person doing changes to CI to test that everything works fine. I doubt we'll be changing it a lot in the future, a decomp project has a finite lifespan.

Mrkol (changes requested):

tools/test_symbol_validation.py
fjooord: Yea I meant to open up as draft first, currently removing these tests.

Clankers love to test lol

fjooord:
Included the whitespace-only fix in `include/JSystem/JAudio/JALibrary/JALList.hpp`, the only file failing the full-tree formatting check.

The exact workflow command now passes locally across all 1,475 tracked source files in one Docker container, taking 15.7 seconds with the image cached. Including this fix should let the full-tree formatting job pass when this PR is merged. Remote CI still needs to confirm; this formatting result does not establish build success.

tools/test_symbol_validation.py
fjooord: Removed the test file and its workflow step in df16b6cc. Verification stays local; no CI tests are added to master.

.github/workflows/build.yml
fjooord: Fixed in df16b6cc. A base object with no function symbols now gets a zero-error baseline, so its first implementation must pass strict validation. Verified locally that missing functions, wrong order, and incorrect linkage fail, while a complete valid implementation passes. Existing nonempty baselines still report inherited defects as warnings.

PR #172, opened 2026-09-15
MoveBG/MapObjGeneral: thrown clears the hit flag, not the live flag; reassociate the throw velocity

gitRasheed:
thrown() cleared the wrong flag, the object calls offHitFlag; the throw velocity is also reassociated the way the object multiplies it.


src/MoveBG/MapObjGeneral.cpp
```diff
@@ -187,7 +189,7 @@ void TMapObjGeneral::sinking()
 		}
 	}
 
-	if (mPosition.y + mMapObjData->mHit->unkC[2].unk4 < unk144) {
+	if (mPosition.y + getMapObjData()->mHit->unkC[2].unk4 < unk144) {
```
Mrkol: This is an arbitrary choice of an inline. They might've had an inline for `mMapObjData->mSink->unk0` or any other subexpression in here -- I prefer not to be too hasty with finding such local minima. What I think we should do is make clankers test inlines en masse. If it makes stack sizes work out in many places at once, that improves the probability of an inline being real.

Lets not include this change and merge the others

Mrkol (changes requested):

src/MoveBG/MapObjGeneral.cpp
gitRasheed: yeah makes sense, I pushed the change, was on a good streak of avoiding stack frame BS ☹

PR #178, opened 2026-09-16
MoveBG/MapObjDolpic: match the static initializer and fix the monument's angle arguments

gitRasheed:
TMonumentShine::control had MsAngleDiff's arguments the wrong way round, and TTurboNozzleDoor zeroes its vector in the constructor body the way the object does.


src/MoveBG/MapObjDolpic.cpp
```diff
@@ -125,6 +126,7 @@ BOOL TMonumentShine::receiveMessage(THitActor* sender, u32 message)
 
 void TMonumentShine::control()
 {
+	// TODO: the target frame is 0x30 rather than 0x38.
```
Mrkol: Clanker comment, remove!

Mrkol (changes requested):

src/MoveBG/MapObjDolpic.cpp
gitRasheed: Gone.

PR #180, opened 2026-09-16
GC2D/CardLoad and CardSave: match the score screen, drawMessage and waitForAnyKey

gitRasheed:
The card screens' fabricated message-id getter goes away in favour of the table read the object makes, and ExPane's centred-size helper computes its offsets from the bounds directly, which is what both screens compile against.


src/GC2D/CardLoad.cpp
```diff
@@ -2090,6 +2088,8 @@ void TCardLoad::setSelected(u8 param_1)
 
 void TCardLoad::changeScene()
 {
+	// TODO: instruction stream matches; the frame is 0x1e0 in the original,
+	// 0x178 here.
```
Mrkol: Clanker comment! Remove!

src/GC2D/CardSave.cpp
```diff
@@ -324,8 +324,8 @@ s8 TCardSave::waitForStop(TEProgress param_1)
 
 	switch (unk10) {
 	case 0:
-		setMessage(unkA0, 0x200, getCurMessageID());
-		setMessage(unkA4, 0x200, getCurMessageID());
+		setMessage(unkA0, 0x200, (u16)cMessageID[unk310]);
+		setMessage(unkA4, 0x200, (u16)cMessageID[unk310]);
```
Mrkol: Is this for stack frame size matching, or does it influence codegen? Because if it's for stack frame, might be a local minima again. Like, why would they write these u16 casts manually? A function returning it makes more sense, no?

src/GC2D/CardLoad.cpp
gitRasheed: Gone.

src/GC2D/CardSave.cpp
gitRasheed: Codegen: main's CardLoad already writes `(u16)cMessageID[unk1C]` at two sites and those are exact. With the getter put back, drawMessage keeps the same instructions but the frame grows by 8. Happy to keep the getter if you'd rather lose that one.

src/GC2D/CardLoad.cpp
Mrkol: Still there

PR #181, opened 2026-09-16
Enemy/BathtubKiller: match the params, init, reset, item generation and the death nerves

gitRasheed:
The colour fields are named from the material names the object passes to getIndex, and TBathtub::getNumGripsDead returns int because its caller compares the result with cmpwi.


src/Enemy/BathtubKiller.cpp
```diff
@@ -14,42 +31,63 @@ static const char* bathtubkiller_bastable[] = {
...
 TBathtubKillerPersonality::TBathtubKillerPersonality() { }
 
-void TBathtubKillerPersonality::makeFast(const TBathtubKillerParams*) { }
+void TBathtubKillerPersonality::makeFast(const TBathtubKillerParams* params)
+{
+	unk0  = params->fastAccelerationQuatRate.get();
+	unk4  = params->fastChaseAcceleration.get();
+	unk8  = params->fastChaseSpeed.get();
+	unkC  = params->fastInitialSpeed.get();
+	unk10 = params->fastDeadPeriod.get();
```
Mrkol, line 40: Lets name these fields -- pretty unambiguous what they should be from this

PR #182, opened 2026-09-16
Enemy/Kumokun: match moveObject and prepareFly, fix the manager's damage radius

gitRasheed:
moveObject reaches the crash counter through isCrashing(), which the object reloads inside the branch the way a deferred inline does, and TSolidStack hands back const references so the element is loaded through its address.


src/Enemy/Kumokun.cpp
```diff
@@ -894,7 +894,9 @@ bool TKumokun::isFlying() const
 	return mSpine->getLatestNerve() == &TNerveKumokunFly::theNerve();
 }
 
-bool TKumokun::isCrashing() const { }
+// TODO: mario.MAP gives this 76 bytes, so the body is longer than the guard
+// TKumokun::moveObject inlines, which is only the unk1D0 test.
+bool TKumokun::isCrashing() const { return unk1D0 > 0; }
```
Mrkol, line 899: Size being different from the map means you reconstructed it wrong,  lets not commit clearly incorrect stuff to master

Mrkol (changes requested):

src/Enemy/Kumokun.cpp
gitRasheed: Yeah, isCrashing is UNUSED at 0x4c in the map so there's nothing to check it against. Closing this one, the small fixes can ride with a later Kumokun PR.

PR #190, opened 2026-09-17
Enemy/smallEnemy, Enemy/bgtentacle: match kill and changeStateAndFixNodes

pablomeendez:
## Summary

Matches two enemy functions:

- `TSmallEnemy::kill()`
  - Corrects the dead-state early return.
  - Moves the `LIVE_FLAG_UNK40` update to match the original control flow.

- `TBGTentacle::changeStateAndFixNodes(int)`
  - Replaces fabricated position and node accessor calls with direct source expressions.
  - Restores the target `0x80` stack frame.

## Verification

- `TSmallEnemy::kill()`: 100.0%
- `TBGTentacle::changeStateAndFixNodes(int)`: 100.0%
- Symbol-order validation passes for both modified units.
- `ninja changes_all`: no regressions.


src/Enemy/bgtentacle.cpp
```diff
@@ -770,18 +770,18 @@ void TBGTentacle::changeStateAndFixNodes(int new_state)
 
 	switch (mState) {
 	case 1:
-		getFirstNode()->onUnk24();
+		mNodes[0].onUnk24();
```
gitRasheed, line 773: @Mrkol can probably weigh in better here, but is this another case of chasing the frame? resetAllNodes matches with getFirstNode() and stops matching if you remove it

Mrkol: I fabricated those inlines, but I see that this has way too much stack with them. I'm OK with removing them here

PR #191, opened 2026-09-18
Structural matching: launcher, hamukuri, generator, telesa, MarioMove

KakarottoCake:
## Summary

Structural-bucket matching across six units, all in already-decompiled TUs. 12 functions to 100%, matched_code 41.01% -> 41.07%.

- launcher 25/35 -> 30/35. getSaveParam2 called TSpineEnemy::getSaveParam explicitly, forcing a direct call where retail dispatches virtually. Dropped the qualification to match Amenbo, BathtubKiller and CoasterKiller. init, resetLaunchTimer, stateLaunch, stateDie and TCommonLauncher::stateDie all 100% from the one line.
- hamukuri +4. TBossDangoHamuKuri zeroes its body counter and TDoroHaneKuriManager nulls its hige pointer in the ctors (both match reset and the null check before use; BossDango createEnemyInstance followed). swingBody copies mRotation, not mPosition: the six uniform +32 accesses are 0x30, the header layout is correct.
- MarNameRefGen getNameRef 100% (1863 instructions). Retail news 68 bytes for TGenerator, we allocate 64. Added trailing unk40, marked unobserved (never ctor-init or read in the TU).
- telesa perform 332/348 -> 344/348. Named MtxPtr for the stack matrix, a shape used elsewhere in the tree. Last instruction is float reload versus shuffle residue.
- MarioMove isWallInFront 100%. Named s16 wallAngle local first, the shape MarioRun uses. Measured byte identical before committing.

## Verification

- mario.dol: OK, bare ninja full build.
- validate-symbol-order passes on launcher, hamukuri, generator, telesa, MarioMove, MarNameRefGen.
- review-lint clean, clang-format clean.
- changes_all against a pre-work baseline: only the listed functions move, no regressions.

## Open questions

1. TCameraMapTool copy ctor (91.7%): members at 0x18/0x1C copy as ints in retail, floats in ours, while load and calcPosAndAt match 100% with float uses. Points at the JGeometry TVec2 copy shape, which is JSystem area, so I left it. The header already suspects SMS TVec2 differs from the SMG/TP inlines.
2. TVector<TCameraMapTool>::InsertRaw (60.2%, 200 vs 320 instr): retail has an explicit element loop with a sizeof division; our std-vector.hpp version differs. Same reason for leaving it: JSystem header.
3. Several pools diffs (fireWanwan attackToMario, telesa initAttacker) ride on the known DummyStrings .sdata shift, not on function code. Deferred as cross-TU header work.
4. Investigated and deliberately left: hamukuri moveObject/attackToMario/createEnemyInstance, fireWanwan updateRumble, EventWatcher evGetTalkNPCName/dispatchBuiltin/evStartMareBottleDemo/evSetTransScale, smallEnemy attackToMario/behaveToHitOthers, MarioMove checkGraffito/moveRequest, MapEventMare bump family, DrawUtil clash. Each is a callee-inline decision, allocator coloring, scheduling order or branch folding with identical logic; the spellings that would close them fail the readability filter.


KakarottoCake:
Update: pushed one more commit. TTelesa::initAttacker read mPosition from the mtx[3] row instead of the translation column (mtx[0][3], mtx[1][3], mtx[2][3]). Same wrong-member class as the swingBody fix. Only a demangler artifact remains in the diff.

KakarottoCake:
Pushed three more hamukuri fixes: TDangoHamuKuri::reset rolls TMsRange 0 to 10, the DangoHamuKuriWait nerve rolls 0 to 30 the same way, and TDangoHamuKuri::calcRootMatrix uses a negated TMsRange 0 to 20 plus the translation column instead of the matrix row. THaneHamuKuri::walkBehavior reconstruction continues locally and stays out of this PR until it matches.

PR #194, opened 2026-09-19
System/MarNameRefGen: pass name tags in getNameRef_Map, add SunModel and PollutionTest

KakarottoCake:
## Summary
- `getNameRef_Map` 72.5% -> 91.8%: pass the name tag retail passes at every `new`, add the `SunModel`/`SunsetModel`/`PollutionTest` branches. Tags verified against retail rodata (Shift-JIS decoded): TMapマップ, TMapDrawWallカメラ食い込み壁, TSky空, TShimmer<Shimmer>, TPollutionManager落書き管理, TPollutionTest落書きテスト, TSunMgr<TSunMgr>, TSunModel(false/true, <TSunModel>), TMarineSnowMarineSnow, TStickyStainManager<TViewObj>.
- add the missing `const char*` ctors for `TStickyStainManager` and `TPollutionTest` as header inlines; retail open-codes the TNameRef/TViewObj chain at both call sites.
- triage for the rest of the cluster (same branch, follow-up commits): `getNameRef_Enemy` 54.1% needs 17 dispatch entries retail has and we lack (AmiNoko, BeeHive, BombHei, Cannon, FruitsBoatB/C/D + managers, Killer, Kukku, MoePuku, Pakkun, Popo, Yumbo); our FruitsBoatG/P/Y guesses are wrong, retail uses B/C/D with `TFruitsBoatManager(int, tag)`. `getNameRef_BossEnemy` 35.5% needs 29 entries. Full 140-branch retail dispatch map extracted; most entries also need explicit tags.

## Residual
- two placement-new clusters differ only in the TNameRef chain: retail calls the out-of-line `TNameRef::TNameRef`, ours inlines the header version. JSystem header decision, left for a human.

## Verification
- `ninja build/GMSJ01/src/System/MarNameRefGen_Map.o`: builds
- `tools/decomp-diff.py -u mario/System/MarNameRefGen_Map -d getNameRef_Map`: 91.8%, remaining diff is the two TNameRef chains only
- `tools/check-changed-symbol-order.py src/System/MarNameRefGen_Map.cpp`: 1 passed
- data-pool: rodata MISSING constants resolved, no new extras
- clang-format clean


src/System/MarNameRefGen_Map.cpp
```diff
@@ -16,40 +18,37 @@
 JDrama::TNameRef* TMarNameRefGen::getNameRef_Map(const char* name) const
 {
 	if (strcmp(name, "Map") == 0)
-		return new TMap;
+		return new TMap("マップ");
```
Mrkol, line 21: These japanese names translate to precisely the name of the class, so all of them should be defaulted arguments. E.g. `TMap(const char* name = "マップ")`.  For `<TViewObj>` -- that means that the TStickyStainManager did NOT accept a `const char* name` and didn't pass it to the parent so the parent constructor defaulted to it's own name.

Mrkol (changes requested):

